PR #1, opened 2019-11-01
initial opentracing

runtologist:


src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+    def span(
+        opName: String,
+        tagError: Boolean = true,
+        logError: Boolean = true
+    ): ZIO[R with COT, E, A] =
+      Telemetry.span(zio, opName, tagError, logError)
+
+    def span(
+        span: Span,
+        tagError: Boolean,
```
mijicd: Perhaps add defaults here?

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+  private def setSpan(span: Span): ZIO[Telemetry, Nothing, Unit] =
+    ZIO.accessM[Telemetry](_.telemetry.spanRef.set(span))
+
+  private def getTracer: ZIO[Telemetry, Nothing, Tracer] =
+    ZIO.environment[Telemetry].map(_.telemetry.tracer)
+
+  private def getCurrentTimeMicros: ZIO[Clock, Nothing, Long] =
+    ZIO.accessM[Clock](_.clock.currentTime(TimeUnit.MICROSECONDS))
+
+  implicit class Implicits[R, E, A](zio: ZIO[R, E, A]) extends AnyRef {
```
mijicd: I'd move this one to package object and make it extend `AnyVal`. Perhaps, I'd rename it to `TelemetrySyntax` just to align it with somehow standardized naming.

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+      tracer <- getTracer
+      old <- getSpan
+      child <- UIO(tracer.buildSpan(opName).asChildOf(old).start())
+      r <- span(zio, child, tagError, logError)
+    } yield r
+
+  def span[R, R1 <: R with COT, E, A](
+      zio: ZIO[R, E, A],
+      span: Span,
+      tagError: Boolean,
```
mijicd: Should these have the default value?

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+      zio: ZIO[R, E, A],
+      opName: String,
+      tagError: Boolean = true,
+      logError: Boolean = true
+  ): ZIO[R1, E, A] =
+    getTracer.flatMap { tracer =>
+      Task(tracer.extract(format, carrier))
+        .fold(_ => None, Option.apply)
+        .flatMap {
+          case None => zio
```
mijicd: I'd fold this.

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+import zio.clock.Clock
+
+trait Telemetry extends Serializable {
+  def telemetry: Telemetry.Service
+}
+
+object Telemetry {
+
+  trait Service {
+    def spanRef: FiberRef[Span]
```
mijicd: Perhaps rename to `currentSpan`.

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+        span <- UIO(tracer.buildSpan(rootOpName).start())
+        ref <- FiberRef.make(span)
+        tracer_ = tracer
+      } yield new Telemetry.Service {
+        override def spanRef: FiberRef[Span] = ref
+        override def tracer: Tracer = tracer_
+      }
+    )(_.spanRef.get.flatMap(span => UIO(span.finish)))
+
+  type COT = Clock with Telemetry
```
mijicd: This can probably be private.

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+
+  def managedService(
+      tracer: Tracer,
+      rootOpName: String = "ROOT"
+  ): ZManaged[Clock, Nothing, Telemetry.Service] =
+    ZManaged.make(
+      for {
+        span <- UIO(tracer.buildSpan(rootOpName).start())
+        ref <- FiberRef.make(span)
+        tracer_ = tracer
```
mijicd: Is this one needed?

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+      } yield new Telemetry.Service {
+        override def spanRef: FiberRef[Span] = ref
```
mijicd: These can be vals.

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+}
+
+object Telemetry {
+
+  trait Service {
+    def spanRef: FiberRef[Span]
+    def tracer: Tracer
+  }
+
+  def managedService(
```
mijicd: I'd rename it to either `make` or `managed`.

src/main/scala/zio/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,213 @@
...
+            span(
+              zio,
+              tracer.buildSpan(opName).asChildOf(spanCtx).start,
+              tagError,
+              logError
+            )
+        }
+    }
+
+  def rootSpan[R, R1 <: R with COT, E, A](
```
mijicd: Maybe just `root`? Also, it looks that it is a child of the root created in `managedService`.

src/main/scala/zio/telemetry/Telemetry.scala
runtologist: We can only have one `span` method  with defaults.

src/main/scala/zio/telemetry/Telemetry.scala
runtologist: We can only have one `span` method  with defaults.

src/main/scala/zio/telemetry/Telemetry.scala
runtologist: With fold we need type parameters because they are not inferred correctly. I prefer pattern matching over explicit type parameters. :)

src/main/scala/zio/telemetry/Telemetry.scala
runtologist: Without it, line 37 is compiled into a recursive call.

src/main/scala/zio/telemetry/Telemetry.scala
runtologist: done

src/main/scala/zio/telemetry/Telemetry.scala
runtologist: done

src/main/scala/zio/telemetry/Telemetry.scala
runtologist: With fold we need explicit type parameters. In that case, I prefer pattern matching.

Issue #6, opened 2019-11-02
Example App


runtologist:
We should illustrate use of zio-telemetry using an example app. It should consist of multiple HTTP services depending on each other. The first service should inject the trace context to all requests it makes. The other services extract and propagate that information. While a first version could just log tracing information, it would be nice to spin up a container running an actual opentracing collector, such as Jaeger to collect data.

* [x] Build at least two services
* [ ] Inject / Extract tracing headers
* [x] Log traces
* [x] Collect traces in Jaeger

runtologist:
Jaeger provides an all in one Docker image using in memory storage for testing here: https://www.jaegertracing.io/docs/1.6/getting-started/#all-in-one-docker-image

drmarjanovic:
@bblagojevic94 and I will take all of this.

runtologist:
Awesome. Let me know, if I can help.

PR #19, opened 2019-11-03
Provide example proxy service

drmarjanovic:
This PR resolves one part of #6.


zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/Server.scala
```diff
@@ -0,0 +1,72 @@
...
+  type AppTask[A] = ZIO[Clock, Throwable, A]
+
+  def app(config: ServiceConfig, tracer: Tracer): HttpApp[AppTask] = {
+    implicit def decoder[A: Decoder]: EntityDecoder[AppTask, A] = jsonOf
+
+    implicit def encoder[A: Encoder]: EntityEncoder[AppTask, A] = jsonEncoderOf
+
+    val routes = HttpRoutes.of[AppTask] {
+      case GET -> Root / "statuses" =>
+        JaegerTracing.makeService(tracer).use { env =>
```
runtologist: I think the service should be created once during application startup. AFAIK, this will start a fresh one per request?

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/Server.scala
```diff
@@ -0,0 +1,72 @@
...
+          (for {
+            _ <- env.telemetry.root("/statuses")
```
runtologist: On current master, there are methods on the companion objects of Telemetry and OpenTracing, which allow for tagging and injection without having to extract the environment manually.

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/Server.scala
```diff
@@ -0,0 +1,72 @@
...
+            _ <- env.telemetry.tag("proxy-tag", "proxy-tag-value")
+            _ = env.telemetry.inject(HttpHeadersFormat, new TextMapAdapter(mutable.Map.empty[String, String].asJava))
```
runtologist: What is the motivation for inject here? The textMap is immediately discarded. Btw, It might be niche to add convenience methods e.g. for HttpHeadersFormat, which do not side effecting update a text map, but return a UIO[Map[String, String]].

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/Server.scala
mijicd: Oh yes, it has to be extracted.

mijicd: Side note, I was thinking whether it's a good idea to provide provider-specific factories (in separate JARs), e.g.:

```scala
def makeService[E](f: ProviderSpecificConfig => IO[E, Tracer]): ZManaged[Clock, Throwable, Clock with OpenTracing]
```

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/config/JaegerConfig.scala
```diff
@@ -0,0 +1,3 @@
+package zio.telemetry.example.proxy.config
+
+final case class JaegerConfig(host: String)
```
mijicd: Please make it a value class.

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/config/ServiceConfig.scala
```diff
@@ -0,0 +1,3 @@
+package zio.telemetry.example.proxy.config
+
+final case class ServiceConfig(url: String)
```
mijicd: Same here.

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/http/StatusesResponse.scala
```diff
@@ -0,0 +1,5 @@
+package zio.telemetry.example.proxy.http
+
+final case class StatusResponse(name: String, version: String, status: String)
+
+final case class StatusesResponse(statuses: StatusResponse)
```
mijicd: Based exclusively on naming, I'd presume `statuses` should be `List[StatusResponse]`. I'd probably do the following as well:

- Make `StatusesResponse` value class.
- Drop `Response` suffix.
- Add encoders to their companions.

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/Server.scala
```diff
@@ -0,0 +1,72 @@
+package zio.telemetry.example.proxy
+
+import cats.effect.ExitCode
+import io.circe.generic.auto._
```
mijicd: Don't use this one. Rather define the encoders as suggested below, and use `semiauto` or specify instances explicitly.

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/config/Config.scala
```diff
@@ -0,0 +1,3 @@
+package zio.telemetry.example.proxy.config
+
+final case class Config(api: ApiConfig, service: ServiceConfig, jaeger: JaegerConfig)
```
mijicd: I'd put all of the other configurations into the `Config`'s companion.

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/config/ServiceConfig.scala
drmarjanovic: I've tried but as I can see in PureConfig [documentation](https://pureconfig.github.io/docs/) it won't work properly. 

My intend here is to have configuration in the following format: 
```
service {
  url = "..."
}
``` 
And making it value class would expect format `url = "..."`. 

Check `final case class Port(value: Int) extends AnyVal` and reference [application.conf](https://github.com/pureconfig/pureconfig/blob/master/docs/src/main/resources/application.conf) from docs. 

If there's any way to do it, please let me know.

modules/example/src/main/resources/application.conf
```diff
@@ -0,0 +1,12 @@
+proxy {
```
mijicd: Group under a single namespace, e.g. `proxy`, and **simplify**:

```
proxy {
  host = "0.0.0.0"
  port = 8080
  backend = "..."
  tracer = "..."
}
```

modules/example/src/main/scala/zio/telemetry/example/config/Config.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import zio.telemetry.example.config.Config.{ BackendConfig, ProxyConfig, TracerConfig }
+
+final case class Config(proxy: ProxyConfig, backend: BackendConfig, tracer: TracerConfig)
+
+object Config {
+
+  final case class ProxyConfig(host: String, port: Int)
+
+  final case class BackendConfig(url: String)
```
mijicd: See the comment above regarding the config structure.

modules/example/src/main/scala/zio/telemetry/example/config/Configuration.scala
```diff
@@ -0,0 +1,25 @@
...
+    val load: RIO[R, Config]
+  }
+
+  trait Live extends Configuration {
+    override val config: Service[Any] = new Service[Any] {
+      val load: Task[Config] = Task.effect(ConfigSource.default.loadOrThrow[Config])
+    }
+  }
+
+  object Live extends Live
```
mijicd: Drop it if you don't use it.

modules/example/src/main/scala/zio/telemetry/example/http/StatusesService.scala
```diff
@@ -0,0 +1,45 @@
...
+        service.use { env =>
+          (for {
+            _ <- env.telemetry.root("/statuses")
+            _ <- OpenTracing.tag("proxy-tag", "proxy-tag-value")
+            _ <- OpenTracing.inject(HttpHeadersFormat, new TextMapAdapter(mutable.Map.empty[String, String].asJava))
+            res <- Client
+                    .status(backendUri)
+                    .map(_.body)
+                    .map {
+                      case Right(s) => Response(Ok).withEntity(Statuses(List(s, StatusUp)))
```
mijicd: DSL can simplify significantly, e.g.

```scala
case Right(s) => Ok(Statuses(List(s, StatusUp)))
```

zio-telemetry-example-proxy/src/main/scala/zio/telemetry/example/proxy/config/ServiceConfig.scala
mijicd: See the comments.

modules/example/src/main/scala/zio/telemetry/example/config/Configuration.scala
drmarjanovic: It's used in [ProxyServer](https://github.com/zio/zio-telemetry/pull/19/files#diff-777a26bc95f4a2a5e38d61f3464ce55bR17).

PR #22, opened 2019-11-05
Restructure sbt project

mijicd:
### Summary

- Moves subprojects to `modules`.
- Removes `2.11` support. 

Minor reason for removing 2.11 support was the failing build due to non-supported version of Circe in examples. Major reason is maintenance effort needed for keeping up with old scala version.


build.sbt
```diff
@@ -58,17 +61,17 @@ lazy val `zio-telemetry` =
 
 testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework"))
 
-val circeVersion  = "0.12.3"
```
runtologist: Can we move the version definitions together?

mijicd: Good point, done.

PR #41, opened 2019-11-21
Improve code in `example` module

drmarjanovic:


modules/example/src/main/scala/zio/telemetry/example/http/StatusesService.scala
```diff
@@ -39,24 +39,22 @@ object StatusesService {
                     .status(backendUri.path("status"), headers)
                     .map(_.body)
                     .flatMap {
-                      case Right(s) => Ok(Statuses(List(s, StatusUp)))
-                      case _        => Ok(Statuses(List(Status("backend", "1.0.0", "down"), StatusUp)))
+                      case Right(s) => Ok(Statuses(List(s, Status.up("proxy"))))
```
mijicd: Don't repeat yourself.

modules/example/src/main/scala/zio/telemetry/example/http/Status.scala
```diff
@@ -3,9 +3,13 @@ package zio.telemetry.example.http
 import io.circe._
 import io.circe.generic.semiauto._
 
-final case class Status(name: String, version: String, status: String)
+final case class Status private (name: String, status: String)
```
mijicd: You can leave it public. If you want it to be properly private, more ceremony is needed (disabling copy), but there's no need for that.

PR #119, opened 2020-04-11
Implements Opentelemetry support

### LINKED ISSUE #107: Draft OpenTelemetry module
As of `v0.1.0`, OpenTelemetry abstractions are temporary dropped to simplify migration to new ZIO version. All future versions should support both OpenTracing and OpenTelemetry, preferrably as separate modules. 

The implementer of this ticket should design an OpenTelemetry module, around the existing [JVM client](https://github.com/open-telemetry/opentelemetry-java/releases).

arjun-1:
Fixes #107 
Hi everyone :slightly_smiling_face: ,

I've worked on a little project to provide support for [OpenTelemetry](https://opentelemetry.io/). I took insipration from the OpenTracing module, and hoped to stay close to its spirit.

This PR includes:
* Updated documentation, including how to run an example using OpenTelemetry.
* Tests. All tests have been reused from OpenTracing. All the tests for OpenTelemetry (but 1) have the same inputs and expectations as for OpenTracing. They are almost an identical copy.
* An example application which can be run as referred to in the documentation (under `modules/example/src/main/scala/zio/telemetry/opentelemetry/`). This is almost an identical copy from the OpenTracing example, and contains therefore some duplicated code. I didn't take time to refactor this common code out, but would happily do so if it is believed to be necessary for this PR 

I would be very happy to hear all of your suggestions and comments to improve this PR :slightly_smiling_face:.

The one notable difference with the OpenTracing module (being the only different test expectation from OpenTracing as well) is the fact that OpenTelemetry doesn't initialize with a 'fake' span that is started when instantiating the layer. It uses a 'dummy' span which does not start instead. This is more in line with the original Java API (which also doesn't start a span when initializing a tracer)

Screenshots of the to the Jaeger UI that the tracing is working as intended:

Screenshot from 2020-04-11 09-29-16

Screenshot from 2020-04-11 10-08-45


runtologist:
Wow, thanks a lot for this contribution, @arjun-1! 🙏  I'll do a review over the weekend.

modules/opentelemetry/src/main/scala/zio/opentelemetry/CurrentSpan.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * Sets the current span.
+   */
+  def setCurrentSpan(span: Span): ZIO[OpenTelemetry, Nothing, Unit] = currentSpan.flatMap(_.set(span))
+
+  /**
+   * Extracts the span from carrier `C` and set its child span with name 'spanName' as the current span.
+   */
+  def createChildFromExtracted[C](
```
arjun-1, line 30: Upon reading the documentation more, I realized that we probably should not expose any functionality to the end-user, that allows setting the span of the `FiberRef` without reverting it to the 'old' span. That probably this function, together with all other functions that set the span in the `FiberRef` (except those in `SpanSyntax`) should be hidden from the user?

modules/opentelemetry/src/main/scala/zio/opentelemetry/OpenTelemetry.scala
```diff
@@ -0,0 +1,81 @@
...
+import zio.clock.Clock
+import zio.{ FiberRef, UIO, URIO, URLayer, ZIO, ZLayer }
+import zio.opentelemetry.SpanUtils.endSpan
+
+object OpenTelemetry {
+  trait Service {
+    val currentSpan: FiberRef[Span]
+    def createRoot(spanName: String, spanKind: Span.Kind): UIO[Span]
+    def createChildOf(spanName: String, parent: Span, spanKind: Span.Kind): UIO[Span]
+    def createChild(spanName: String, spanKind: Span.Kind): UIO[Span]
```
arjun-1, line 15: This probably does not belong here and should go to `CurrentSpan`, as it is not an elementary operation build on `Tracer`

modules/opentelemetry/src/main/scala/zio/opentelemetry/OpenTelemetry.scala
```diff
@@ -0,0 +1,81 @@
+package zio.opentelemetry
+
+import java.util.concurrent.TimeUnit
+
+import io.opentelemetry.trace._
+import zio.clock.Clock
+import zio.{ FiberRef, UIO, URIO, URLayer, ZIO, ZLayer }
+import zio.opentelemetry.SpanUtils.endSpan
+
+object OpenTelemetry {
```
arjun-1, line 10: Given that OpenTelemetry also supports metrics: https://github.com/open-telemetry/opentelemetry-java/blob/master/QUICKSTART.md#metrics, this might better be called `Tracer`?

In the future, we could then have 2 modules, each named after the provider on the java OpenTelemetry:
`getMeterProvider` -> `zio.opentelemetry.Meter`
`getTracerProvider` -> `zio.opentelemetry.Tracer`

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/SpanUtils.scala
```diff
@@ -0,0 +1,22 @@
...
+  /**
+   * Sets the status of `span` to `UNKNOWN` error with description being the pretty-printed cause.
+   */
+  def setErrorStatus(span: Span, cause: Cause[_]): UIO[Unit] =
+    UIO(span.setStatus(Status.UNKNOWN.withDescription(cause.prettyPrint)))
+
+  /**
+   * Ends the span with the current time.
+   */
+  def endSpan(span: Span): URIO[Clock, Unit] = {
```
arjun-1, line 18: Would we want to have this function inside `Tracing.Service`? Right now, the consumers of this library would need to provide `Clock` twice (as shown in the example):
```scala
Clock.live ++ (Clock.live >>> Tracing.live(tracer))
```
1 time for determining starting timestamps, another for ending timestamps. By having `endSpan` in `Tracing.Service`, that would be remedied.

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/ContextPropagation.scala
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Extract and returns the span from carrier `C`.
+   */
+  def extractSpan[C](
+    httpTextFormat: HttpTextFormat,
+    carrier: C,
+    reader: PropagationFormat.Reader[C]
+  ): UIO[Span] =
+    semaphore.flatMap(_.withPermit {
+      UIO(httpTextFormat.extract(Context.ROOT, carrier, getter(reader))).map(TracingContextUtils.getSpan)
```
runtologist, line 51: I think this needs to be uninterruptible. Otherwise, even with the semaphore, something might concurrently set a different span on the thread this gets scheduled on, or the getSpan might get executed on a different thread. That would even work without the Semaphore.

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/SpanUtils.scala
```diff
@@ -0,0 +1,22 @@
...
+import io.opentelemetry.trace._
+import zio.clock.Clock
+import zio.{ Cause, UIO, URIO }
+
+private[opentelemetry] object SpanUtils {
+
+  /**
+   * Sets the status of `span` to `UNKNOWN` error with description being the pretty-printed cause.
+   */
+  def setErrorStatus(span: Span, cause: Cause[_]): UIO[Unit] =
```
runtologist, line 12: How about adding the capability to set a custom error status?

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/TracingSyntax.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  implicit final class OpenTelemetryZioOps[R, E, A](val effect: ZIO[R, E, A]) extends AnyVal {
+
+    /**
+     * @see [[Tracing.spanFrom]]
+     */
+    def spanFrom[C](
+      httpTextFormat: HttpTextFormat,
+      carrier: C,
+      reader: PropagationFormat.Reader[C],
```
runtologist, line 20: Generally speaking, I like the idea of having functional Reader and Writer. However, I guess the Java API is meant to be used with a ready made Reader/Writer from a library such as Jaeger. Defining our own trait here forces the user of the library to implement it. Therefore I propose to stick to the Java API here. We could provide convenience methods for particular formats in a later PR, which completely hide the Java API.

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/SpanUtils.scala
arjun-1: That's a good suggestion

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/TracingSyntax.scala
arjun-1: Good point, I hadn't thought of that use case yet. Indeed it makes sense to follow the Java API here

runtologist:
Two more comments regarding names and structure:

The package and module structure is a bit compilcated now. I propose to rename the `core` module to `opentracing` and introduce `opentelemetry` besides it and analogously have `opentracing-example` and `opentelemetry-example`. Also the package structure should be roughly the same in all modules, e.g. with a `zio.telemetry.(opentelemetry|opentracing)[.example]` prefix. Feel free to move the existing examples. When the opentelemetry examples are in a separate package, we also don't need a custom name for the second `application.conf`.

IMO method names should be roughly the same between `opentelemetry` and `opentracing`. Currently we have different names, e.g. `childSpan` vs `span` and `rootSpan` vs `root`.

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/SpanUtils.scala
runtologist: Absolutely, let's do this. Changing opentracing as you propose has been on the TODO list for a few weeks, but we never got around to doing it. `Clock` should only be needed when constructing the layer. We can adjust opentracing analogously in a separate PR.

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/ContextPropagation.scala
```diff
@@ -0,0 +1,70 @@
...
+  //  2) then does the actual injection of the retrieved span into the carrier, without needing any context
+  //
+  //  Thus we can use a 'dummy' context to play the role of medium: `Context.Root`, which doesn't have any thread local storage.
+  //  However, given there is only a single instance of `Context.Root`, we prevent any synchronization issues by
+  //  permitting access only to a single fiber at a time.
+  //
+  // See https://github.com/open-telemetry/opentelemetry-java/issues/575, and
+  // https://github.com/open-telemetry/opentelemetry-java/issues/1104
+
+  private val semaphore: UIO[Semaphore] = Semaphore.make(1)
```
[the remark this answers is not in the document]
arjun-1: Your comment left me wondering whether we need the `Semaphore` at all. I wrote a small test that tests injection/extraction of a span using `foreachPar` (adapted from round trip test), and indeed the `Semaphore` made no difference (it succeeds anyway).

So indeed a valid point :slightly_smiling_face:

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/TracingSyntax.scala
arjun-1: Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/70dadc084f45d3e1f4b548d40c5486a71d477edd

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/ContextPropagation.scala
arjun-1: Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/85286fb80663bdd87a20d4556812c37d657da7ec

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/ContextPropagation.scala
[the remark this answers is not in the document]
arjun-1: Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/85286fb80663bdd87a20d4556812c37d657da7ec

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/SpanUtils.scala
arjun-1: Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/2537dae04e0fb851317f3c076ff555bc526a501a

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/SpanUtils.scala
arjun-1: Maybe use a partial function which maps the error `E` into a `Status`, so users can decide themselves which particular kind of their error `E` belongs to which status?

For example
```scala

def spanFrom[C, R, E, A](
    httpTextFormat: HttpTextFormat,
    carrier: C,
    getter: HttpTextFormat.Getter[C],
    spanName: String,
    spanKind: Span.Kind = Span.Kind.INTERNAL,
    toErrorStatus: PartialFunction[E, Status] = Map.empty
  )

private def setErrorStatus[E](span: Span, cause: Cause[E], toErrorStatus: PartialFunction[E, Status]): UIO[Unit] = {
    val errorStatus: Status = cause.failureOption.flatMap(toErrorStatus.lift).getOrElse(Status.UNKNOWN)
    UIO(span.setStatus(errorStatus.withDescription(cause.prettyPrint)))
  }
```

runtologist: That's a good idea!

arjun-1: Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/8fdf9fbc403aba42e656abda526196700dc99fb7

modules/opentelemetry/src/main/scala/zio/opentelemetry/tracing/ContextPropagation.scala
arjun-1: Fixed with https://github.com/zio/zio-telemetry/commit/85286fb80663bdd87a20d4556812c37d657da7ec

I didn't know about this feature yet, but looks like it might solve the problem. Ideally we would have a test that dictates having this feature. But given that this problem is likely to be resolved in the next release of Opentelemetry-java (https://github.com/open-telemetry/opentelemetry-java/issues/575), we will not need this soon anymore anyway

arjun-1:
Hi @runtologist. Regarding your comment https://github.com/zio/zio-telemetry/pull/119#issuecomment-613925284

> The package and module structure is a bit compilcated now. I propose to rename the `core` module to `opentracing` and introduce `opentelemetry` besides it and analogously have `opentracing-example` and `opentelemetry-example`. 

I'm all in favor of a simpler and flat structure. Your proposal would look as follows right?
```
.
+-- opentelemetry
|   +-- src
|   +-- ...
+-- opentracing
|   +-- src
|   +-- ...
+-- opentelemetry-example
|   +-- src
|   +-- ...
+-- opentracing-example
|   +-- src
|   +-- ...
+-- build.sbt
+-- ...
```

> Also the package structure should be roughly the same in all modules, e.g. with a `zio.telemetry.(opentelemetry|opentracing)[.example]`

Agree on the equal package structure for the examples as well. Given that opentelemetry also has metrics capabilities, probably this warrants an extra package? Like so:
opentelemetry:
```
zio.opentelemetry.tracing
zio.opentelemetry.metrics          (future)
zio.opentelemetry.tracing.example
zio.opentelemetry.metrics.example  (future)
```
opentracing:
```
zio.opentracing
zio.opentracing.example
```

> IMO method names should be roughly the same between `opentelemetry` and `opentracing`. Currently we have different names, e.g. `childSpan` vs `span` and `rootSpan` vs `root`

Indeed we should aim for the same naming between the modules. As a newcomer I tried to make an overview of the names we have right now, and some proposals I could think of which could improve:
```
zio-telemetry	<-> 	java-opentelemetry
------------------------------------------
current:
  root 		<-> 	buildSpan/start
  span 	        <-> 	buildSpan/asChildOf/setParent/start
proposals:
  root					
  child
  ---
  rootSpan
  childSpan
current:
  inject        <->	inject
  spanFrom      <-> 	extract
proposals:
  injectSpanInto
  extractSpanFrom
  ---
  spanInto
  spanFrom
  ---
  injectInto
  extractFrom
  ---
  inject
  extract
```
What do you think? Naming is hard, so feel free to tell me if any of this doesn't make sense. I will revert to the old existing naming scheme then.

arjun-1:
> Two more comments regarding names and structure:
> 
> The package and module structure is a bit compilcated now. I propose to rename the `core` module to `opentracing` and introduce `opentelemetry` besides it and analogously have `opentracing-example` and `opentelemetry-example`. Also the package structure should be roughly the same in all modules, e.g. with a `zio.telemetry.(opentelemetry|opentracing)[.example]` prefix. Feel free to move the existing examples. When the opentelemetry examples are in a separate package, we also don't need a custom name for the second `application.conf`.
> 
> IMO method names should be roughly the same between `opentelemetry` and `opentracing`. Currently we have different names, e.g. `childSpan` vs `span` and `rootSpan` vs `root`.

I've applied all of your suggestions :slightly_smiling_face:

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
```diff
@@ -0,0 +1,197 @@
...
+            assert(child)(
+              isSome(
+                hasField[SpanData, SpanId](
+                  "parentSpanId",
+                  _.getParentSpanId,
+                  equalTo(root.get.getSpanId)
+                )
+              )
+            )
+        }.provideCustomLayer(createMockLayer),
```
runtologist, line 70: `.provideCustomLayer(createMockLayer)` could be moved to the `suite` level in order to save a few repetitions.

opentelemetry-example/src/main/resources/application.conf
```diff
@@ -0,0 +1,12 @@
+proxy {
+  host = "0.0.0.0"
```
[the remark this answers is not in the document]
runtologist: I noticed the opentracing example does the same. Feel free to leave that, I can clean it up.

runtologist:
@mijicd Do we need any changes for publishing, or should that work out of the box?

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/config/Configuration.scala
```diff
@@ -0,0 +1,24 @@
...
+import zio.ZIO
+
+object Configuration {
+
+  trait Service {
+    val load: Task[Config]
+  }
+
+  object Live extends Service {
+    val load: Task[Config] =
```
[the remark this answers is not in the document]
mijicd: One thing to note here: at the time we wrote the lib, zio-config didn't exist. I will open an issue to replace pureconfig with it.

mijicd:
@arjun-1 many thanks for this contribution!

@runtologist Everything's ready for new artifact to be published.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/ContextPropagation.scala
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Injects the span into carrier `C`.
+   */
+  def injectSpan[C](
+    span: Span,
+    httpTextFormat: HttpTextFormat,
+    carrier: C,
+    setter: HttpTextFormat.Setter[C]
+  ): URIO[Tracing, Unit] =
+    ZIO.uninterruptible {
```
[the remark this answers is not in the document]
arjun-1: You are right. Looking at the discussion https://github.com/open-telemetry/opentelemetry-java/issues/1104 , there is nothing tricky about using `Context.ROOT`

Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/c78e3ddb614789840e40142f86e3cd22ecc17554

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/config/Configuration.scala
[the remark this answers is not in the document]
arjun-1: Your comment incited me to do some refactoring in the opentelemetry example, but the following piece of code actually made me stumble:
```scala
def live(tracer: Tracer): URLayer[Clock, Tracing] 
```
and equivalently in opentracing 
```scala
def live(tracer: Tracer, rootOperation: String = "ROOT"): URLayer[Clock, OpenTracing]
```
Please correct me if I'm wrong, but by using this pattern aren't we actually making it impossible to use this library whenever the user provides a `Tracer` that is an effect (e.g. `Task[Tracer]`) or that is a layer (`ULayer[Has[Tracer]`)? Given there is no `flatMap` on `ZLayer`, I don't see a possibility.

Would the above use cases be made possible again if we would have something like:
```scala
def live: URLayer[Clock with Has[Tracer], Tracing] 
```
instead?

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/ContextPropagation.scala
```diff
@@ -35,9 +35,7 @@ private[opentelemetry] object ContextPropagation {
     carrier: C,
     getter: HttpTextFormat.Getter[C]
   ): UIO[Span] =
-    ZIO.uninterruptible {
-      UIO(httpTextFormat.extract(Context.ROOT, carrier, getter)).map(TracingContextUtils.getSpan)
-    }
+    UIO(httpTextFormat.extract(Context.ROOT, carrier, getter)).map(TracingContextUtils.getSpan)
```
runtologist, line 38: This one must be uninterruptible as it sets and then gets a ThreadLocal. Therefore it absolutely must run on the same thread without anyone touching the ThreadLocal in between.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/ContextPropagation.scala
```diff
@@ -48,11 +46,6 @@ private[opentelemetry] object ContextPropagation {
     carrier: C,
     setter: HttpTextFormat.Setter[C]
   ): URIO[Tracing, Unit] =
-    ZIO.uninterruptible {
-      for {
-        context <- UIO(TracingContextUtils.withSpan(span, Context.ROOT))
-        _       <- UIO(httpTextFormat.inject(context, carrier, setter))
-      } yield ()
-    }
+    UIO(TracingContextUtils.withSpan(span, Context.ROOT)).map(httpTextFormat.inject(_, carrier, setter))
```
runtologist, line 49: This one needs not be uninterruptible AFAIKS, as the inject uses the return value of `withSpan` instead of reading a ThreadLocal.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/ContextPropagation.scala
arjun-1: It must have been really late last night :sweat_smile: , as I discover now I misread your comment and accidentally removed this `uninterruptible`. Sorry for the inconvenience :slightly_smiling_face: , should be fixed again

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
arjun-1: Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/3508773a746bb0c3d5d6722294ccdbea981e601d

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/ContextPropagation.scala
arjun-1: Responded with https://github.com/zio/zio-telemetry/pull/119#discussion_r412207764

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/config/Configuration.scala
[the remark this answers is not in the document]
arjun-1: Fixed single validation of configuration with https://github.com/zio/zio-telemetry/pull/119/commits/3c89f364f50b12e8cf426fd19b26ad0ed03ac3ab

opentelemetry-example/src/main/resources/application.conf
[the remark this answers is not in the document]
arjun-1: Fixed with https://github.com/zio/zio-telemetry/pull/119/commits/3c89f364f50b12e8cf426fd19b26ad0ed03ac3ab

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/config/Configuration.scala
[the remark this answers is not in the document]
runtologist: For opentracing we already have a package private managed. We could make it public in a separate PR.

arjun-1:
NP. Thanks a lot for your time reviewing :slightly_smiling_face:

Issue #130, opened 2020-04-22
OpenTracing module reports 2 traces for a single request


arjun-1:
Currently the opentracing module will report 2 traces for a single (1) request:
![Screenshot from 2020-04-22 10-56-47](https://user-images.githubusercontent.com/8102654/79962296-289cba00-8488-11ea-97ff-55813bd3eaf7.png)

Here, the 'ROOT' span belonging only to 'zio-backend' was merely created when instantiating the module, and is probably of no interest to the user. The only correct one is the second reported trace (having 2 spans) it seems?

When implementing https://github.com/zio/zio-telemetry/pull/119, the same behaviour was encountered at first. It was solved by instantiating the module with a noop, default span.

There exists also a noop span for OpenTracing, and the same approach as OpenTelemetry will probably resolve the issue. Note that it will also solve the burden of choice the user has right now when choosing a rootspan when instantiating the module. (Even though a default name is provided, naming of the root span is probably not something that should be of relevance when instantiating the layer)

arjun-1:
After investigating a possible solution using the noop span, it turns out that this indeed removes the redundant reported trace. It is unusable though, since any use of this noop span as root span (without setting it with a different 'real' span) as a parent span will cause an exception. The reason this does work in opentelemetry is because of differences between the libraries: https://github.com/opentracing/opentracing-java/issues/375

runtologist:
There aren't two spans per trace. The root span lives for the entire lifespan of the application, not per request. We chose this approach for the reason you just described. Does that make sense?

arjun-1:
Ah, thanks for the clarification. Yeah, one span per lifespan of the application would make sense as an approach. Although the aforementioned ROOT spans seem to show up per request, i.e. 1 request => 1 extra reported ROOT span:

![Screenshot from 2020-04-26 09-29-06](https://user-images.githubusercontent.com/8102654/80300898-b803df00-87a0-11ea-8ce8-85cc1c459dbd.png)

These reported ROOT spans have different SpanID's. Or could I be misinterpreting this data?

runtologist:
Then we should take a look at the example apps. Maybe they have a bug. I can take a look next week.

Fristi:
I can confirm this behavior

All root traces (104)

<img width="1657" alt="Screenshot 2020-04-28 at 16 41 00" src="https://user-images.githubusercontent.com/141265/80500937-20a0b680-896f-11ea-9af1-26f202b38d19.png">

All traces (228)
[...]

PR #136, opened 2020-05-02
Introduce UManaged to model a span being a resource in Telemetry module

arjun-1:
As suggested here https://github.com/zio/zio-telemetry/issues/130, indeed Spans can be seen as resources and in that case they can and should be modeled using a `Managed`

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
@@ -9,6 +9,7 @@ import zio.clock.Clock
 import zio.telemetry.opentelemetry.attributevalue.AttributeValueConverter.toAttributeValue
 import zio.telemetry.opentelemetry.SpanPropagation.{ extractSpan, injectSpan }
 import zio._
+import zio.UManaged
 import zio.telemetry.opentelemetry.attributevalue.AttributeValueConverter
 
 import scala.jdk.CollectionConverters._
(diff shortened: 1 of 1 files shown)


opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
```diff
@@ -17,8 +18,9 @@ object Tracing {
   trait Service {
     private[opentelemetry] def currentNanos: UIO[Long]
     private[opentelemetry] val currentSpan: FiberRef[Span]
-    private[opentelemetry] def createRoot(spanName: String, spanKind: Span.Kind): UIO[Span]
-    private[opentelemetry] def createChildOf(parent: Span, spanName: String, spanKind: Span.Kind): UIO[Span]
+    private[opentelemetry] def createRoot(spanName: String, spanKind: Span.Kind): UManaged[Span]
+    private[opentelemetry] def createChildOf(parent: Span, spanName: String, spanKind: Span.Kind): UManaged[Span]
+    private[opentelemetry] def end(span: Span): UIO[Unit]
```
runtologist, line 23: If Spans are managed, why do we need to expose `end`?

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
```diff
@@ -27,29 +29,26 @@ object Tracing {
...
   }
 
   /**
-   * Sets the `currentSpan` to `span` only while `effect` runs.
-   * Then ends `span` according to the result of `effect`.
+   * Sets the `currentSpan` to `span` only while `effect` runs,
+   * and error status of `span` according to any potential failure of effect.
    */
-  private def finalizeSpanUsingEffect[R, E, A](
+  private def setSpanError[R, E, A](
```
runtologist, line 51: Nit: As this does not only set an error status, I liked the previous name better.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
arjun-1: Good point! We don't need to indeed

Fixed with https://github.com/zio/zio-telemetry/pull/136/commits/5189adc8ba1a577ab76892b81bfa834ce9c70551

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
arjun-1: No problem :). Fixed with https://github.com/zio/zio-telemetry/pull/136/commits/1d528cedd20b3bd7fc688cc7d7d58ec429e7bdaf

PR #151, opened 2020-05-18
Upgrade ZIO to 1.0.0-RC19

ghostdogpr:


opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
```diff
@@ -155,16 +155,18 @@ object TracingTest extends DefaultRunnableSpec {
           val log =
             for {
               _ <- UIO.unit.addEvent("message")
+              fiber <- ZIO
+                        .sleep(duration)
```
runtologist, line 159: This sleep was a workaround. it should be possible to simplify `log` to 
```scala
TestClock.adjust(duration)
  .addEvent("message")
  .addEventWithAttributes(
    "message2",
    Map(
      "msg"  -> AttributeValue.stringAttributeValue("message"),
      "size" -> AttributeValue.longAttributeValue(1)
    )
  )
```

The comment above also is resolved then.

opentracing/src/test/scala/zio/telemetry/opentracing/OpenTracingTest.scala
```diff
@@ -139,9 +139,10 @@ object OpenTracingTest extends DefaultRunnableSpec {
            */
           val log =
             for {
-              _ <- UIO.unit.log("message")
-              _ <- TestClock.adjust(duration)
-              _ <- ZIO.sleep(duration).log(Map("msg" -> "message", "size" -> 1))
+              _     <- UIO.unit.log("message")
```
runtologist, line 142: same here

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
ghostdogpr: Hmm if I do that, the test fails because we get `(0,message,Map())` for the first message, instead of `(1000000,message,Map())`, which makes sense. Should I change the expectation?

ghostdogpr: Otherwise I change it to:
```scala
ZIO.unit.addEvent("message") *>
  TestClock
    .adjust(duration)
    .addEventWithAttributes(
      "message2",
      Map(
        "msg"  -> AttributeValue.stringAttributeValue("message"),
        "size" -> AttributeValue.longAttributeValue(1)
      )
    )
```
Which one is better?

runtologist: Ah, right. I think the one you propose is better.

ghostdogpr: Okay, I fixed both.

runtologist:
Thanks for upgrading the ecosystem. 🙏

ghostdogpr:
@runtologist Thanks 😄 Can you do a release? If not, let me know and I can do it. I'll upgrade sttp next.

runtologist:
@ghostdogpr I can do a release.

PR #169, opened 2020-06-18
Fix nexus link and make dependency explicit

ghostdogpr:


mijicd:
@ghostdogpr seems like we're facing an issue with jdk11 similar to the one on core.

ghostdogpr:
Yeah. Maybe we can merge that since it’s only touching the ReadMe.

mijicd:
Agreed. I'll have some time later to see what's going on with this.

PR #190, opened 2020-07-16
OpenTracing layer from entry point

Fristi:
I've introduced a variant to `OpenTracing.live` named `OpenTracing.fromEntrypoint`. 

This takes two parameters:

- A `FiberRef[A]` which can scoped for each request by using `FiberRef.locally`
- A `EntrypointExtractor[A]` which extracts the current scoped value from the `FiberRef` and turns in to a `Entrypoint` which has a name for the `Span` and a `SpanContext` for propagating context across async boundaries.

With the introduction of this `OpenTracing.fromEntrypoint` each time this dependency is requested it needs to be created each time. Therefore the use of `.fresh`

I've refactored the example application which illustrates how this works. The `A` is set to `final case class HttpOperation(operationName: String, headers: org.http4s.Headers)`. The companion object implements a `EntrypointExtractor` which uses the `TextMapAdapater` and `tracer.extract` methods to extract a `SpanContext`. 

Around each endpoint in the server you are required to set the context of that request by callling `ref.locally(HttpOperation("<name>", req.headers)`. This can be easily abstracted away if you are using a library like endpoints4s or tapir which might include a operationId and has the request at that point.

If things can be done in a similiar way I am glad to hear, but I think this is a good way of creating contextual root spans by using a custom `FiberRef` and `EntrypointExtractor`

The basic idea is that you have a local view of the `FiberRef` and use `provideLayer` inside that call to construct the `OpenTracing` layer dependency such that every call has it's own entrypoint. Psuedo code:

```scala
ref.locally {
  prg.provideLayer(env)
}
```


runtologist:
Sounds great, thanks for this contribution. I'm on vacation for two weeks, so reviews will be slow.

Fristi:
Enjoy your holidays! Looking forward to your feedback

Fristi:
We could also chose to get rid of the `EntrypointExtractor` and use `FiberRef[Entrypoint]` instead as it's a form indirection

Fristi:
I'll be on holidays later this week for a few weeks, just so you know :)

opentracing/src/main/scala/zio/telemetry/opentracing/OpenTracing.scala
```diff
@@ -26,54 +26,95 @@ object OpenTracing {
...
+    entrypointExtractor: EntrypointExtractor[A]
+  ): URLayer[Clock, OpenTracing] =
+    ZLayer.fromManaged(managedFromEntrypoint(tracer, refEntrypoint, entrypointExtractor)).fresh
+
   def live(tracer: Tracer, rootOperation: String = "ROOT"): URLayer[Clock, OpenTracing] =
     ZLayer.fromManaged(managed(tracer, rootOperation))
 
+  def managedFromEntrypoint[A](
+    tracer0: Tracer,
+    refEntrypoint: FiberRef[A],
```
runtologist, line 57: I'm not sure I understand why we need to provide `refEntrypoint` here. Can't we just create a fresh one inside `managedFromEntrypoint`? Ha ving to provide a `FiberRef` feels a bit low level.

Fristi: The idea is to create the layer with an external mutually variable, hence a `FiberRef`. Upon creation it will extract the current context which set typically by calling `.locally`. This happens when you process a HTTP request or a Kafka message.

PR #208, opened 2020-08-12
Add scoped effects to Tracing service.

### LINKED ISSUE #207: Propagate context into non-zio code
Apologies in advance if this doesn't make sense, I'm still making sense of the opentelemetry world...

I've been playing with the opentelemetry auto instrumentation and zio-telemetry and have both of them submitting spans to jaeger very nicely. In case where requests come in through akka http and play, the auto-instrumentation creates the root span and my zio runtime picks it up successfully. However, my zio code then has to call through into non-zio code (slick, in this case) to access the database. The auto-instrumentation instruments jdbc, which works, but the spans for those sql calls are not connected to the zio telemetry spans.

I believe the issue here is that Tracing.live picks up the default current context, but zio doesn't push a scope onto the thread local context prior to executing the "legacy" code. I'm still trying to figure out how to make this connection...

From what I can tell so far, there a few general approaches we could take to this:

1. Expose the span and eventually document how to propagate the context by hand if desired. This is the simplest for zio-telemetry in that it just means making `currentSpan` public. 

2. Set the thread-local context prior to executing spanned effects. I've been experimenting with this in combination with (1) but don't quite have it working yet. Notionally, it looks something like this:

```
for {
  currentSpan <- Tracing.currentSpan
  result <- ZIO.effect { 
     val scope = Tracer.withSpan(currentSpan)
     try { 
     .... do stuff
     } finally {
        scope.finish()
     }
  }
} yield result
```

Assuming this can be made to work, it seems like it ought to be possible to wire into `Tracing.span()` and avoid the need for exposing `currentSpan` as public.

3. Integrate directly with the various future/callback context propagation systems in play ([otel akka](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/akka-context-propagation/src/main/java/io/opentelemetry/auto/instrumentation/akkaconcurrent/AkkaForkJoinTaskInstrumentation.java), [otel java-concurrent](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/instrumentation/java-concurrent/src/main/java/io/opentelemetry/auto/instrumentation/javaconcurrent/FutureInstrumentation.java), [opentracing scala-concurrent](https://github.com/opentracing-contrib/scala-concurrent)). This seems like it might be the cleaner way to avoid any propagation issues with respect to fibers vs thread pools and such but doesn't necessarily cover the case of directly executed code. Unless there's an issue with using the thread-local context as the conduit, that seems like the obvious solution in that the various concurrency handlers below are designed to pick up their context from the calling thread.

4. Something else entirely?

easel:
By introducing `Tracing.scopedEffect`, `Tracing.scopedEffectTotal`
and `Tracing.scopedEffectFromFuture` it is now possible to pass spans
reliably into non-zio code, such as that generated by the java auto
instrumentation.
    
`Tracing.scopedEffectFromFuture` ensure the scope is preset at the time
the future is created, but does not currently wrap the Future executionContext 
ensuring that the context is passed since the java auto instrumentation
already handles this and it is generally the concern of the `legacy` code
rather than zio-telemetry

Fixes #207

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
@@ -2,6 +2,8 @@ package zio.telemetry.opentelemetry
 
 import java.util.concurrent.TimeUnit
 
+import scala.concurrent.{ ExecutionContext }
+
 import io.opentelemetry.common.Attributes
 import io.opentelemetry.context.propagation.HttpTextFormat
 import io.opentelemetry.trace.{ DefaultSpan, EndSpanOptions, Span, Status, Tracer }
@@ -17,6 +19,7 @@ object Tracing {
     private[opentelemetry] val currentSpan: FiberRef[Span]
     private[opentelemetry] def createRoot(spanName: String, spanKind: Span.Kind): UManaged[Span]
     private[opentelemetry] def createChildOf(parent: Span, spanName: String, spanKind: Span.Kind): UManaged[Span]
+    private[opentelemetry] def getTracer: UIO[Tracer]
   }
 
   private def currentNanos: URIO[Tracing, Long] =
@@ -32,6 +35,9 @@ object Tracing {
     ZManaged.accessManaged[Tracing](_.get.createChildOf(parent, spanName, spanKind))
 
   private def getCurrentSpan: URIO[Tracing, Span] = currentSpan.flatMap(_.get)
+
+  private def getTracer: URIO[Tracing, Tracer] =
+    ZIO.access[Tracing](_.get).flatMap(_.getTracer)
 
   private def toEndTimestamp(time: Long): EndSpanOptions = EndSpanOptions.builder().setEndTimestamp(time).build()
 
@@ -97,6 +103,62 @@ object Tracing {
       old <- getCurrentSpan
       r   <- createChildOf(old, spanName, spanKind).use(finalizeSpanUsingEffect(effect, _, toErrorStatus))
     } yield r
+
+  /*
+   * Introduces a thread-local scope during the execution allowing
+   * for non-zio context propagation.
+   *
+   * Closes the scope when the effect finishes.
+   */
+  def scopedEffect[R, A](effect: => A): ZIO[Tracing, Throwable, A] =
+    for {
+      currentSpan <- getCurrentSpan
+      tracer      <- getTracer
+      eff <- Task.effect {
+              val scope = tracer.withSpan(currentSpan)
+              try effect
+              finally scope.close()
+            }
+    } yield eff
+
+  /*
+   * Introduces a thread-local scope during the execution allowing
+   * for non-zio context propagation.
+   *
+   * Closes the scope when the effect finishes.
+   */
+  def scopedEffectTotal[R, A](effect: => A): ZIO[Tracing, Nothing, A] =
+    for {
+      currentSpan <- getCurrentSpan
+      tracer      <- getTracer
+      eff <- Task.effectTotal {
+              val scope = tracer.withSpan(currentSpan)
+              try effect
+              finally scope.close()
+            }
+    } yield eff
+
+  /*
+   * Introduces a thread-local scope from the currently active zio span
+   * allowing for non-zio context propagation. This scope will only be
+   * active during Future creation, so another mechanism must be used to
+   * ensure that the scope is passed into the Future callbacks.
+   *
+   * The java auto instrumentation package provides such a mechanism out of
+   * the box, so one is not provided as a part of this method.
+   *
+   * CLoses the scope when the effect finishes
+   */
+  def scopedEffectFromFuture[R, A](make: ExecutionContext => scala.concurrent.Future[A]): ZIO[Tracing, Throwable, A] =
+    for {
+      currentSpan <- getCurrentSpan
+      tracer      <- getTracer
+      eff <- ZIO.fromFuture { implicit ec =>
+              val scope = tracer.withSpan(currentSpan)
+              try make(ec)
+              finally scope.close()
+            }
+    } yield eff
 
   /**
    * Injects the current span into carrier `C`
@@ -173,6 +235,9 @@ object Tracing {
                    )
                  )(end)
         } yield span
+
+      override private[opentelemetry] def getTracer: UIO[Tracer] =
+        UIO.succeed(tracer)
 
       val currentSpan: FiberRef[Span] = defaultSpan
     }
--- a/opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
+++ b/opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
@@ -15,9 +15,11 @@ import zio.test.Assertion._
 import zio.test.environment.TestClock
 import zio.test.{ assert, suite, testM, DefaultRunnableSpec }
 import zio._
-
 import scala.collection.mutable
+import scala.concurrent.Future
 import scala.jdk.CollectionConverters._
+
+import io.grpc.Context
 import io.opentelemetry.common.Attributes
 
 object TracingTest extends DefaultRunnableSpec {
@@ -32,13 +34,11 @@ object TracingTest extends DefaultRunnableSpec {
     tracer = tracerProvider.get("TracingTest").asInstanceOf[Tracer]
   } yield (inMemoryTracing, tracer)
 
-  val inMemoryTracerLayer: ULayer[Has[InMemoryTracing] with Has[Tracer]] = ZLayer.fromEffectMany(
-    inMemoryTracer.map {
-      case (inMemoryTracing, tracer) => Has(inMemoryTracing).add(tracer)
-    }
-  )
+  val inMemoryTracerLayer: ULayer[Has[InMemoryTracing] with Has[Tracer]] = ZLayer.fromEffectMany(inMemoryTracer.map {
+    case (inMemoryTracing, tracer) => Has(inMemoryTracing).add(tracer)
+  })
 
-  val tracingMockLayer: URLayer[Clock, Has[InMemoryTracing] with Tracing] =
+  val tracingMockLayer: URLayer[Clock, Has[InMemoryTracing] with Tracing with Has[Tracer]] =
     (inMemoryTracerLayer ++ Clock.any) >>> (Tracing.live ++ inMemoryTracerLayer)
 
   def getFinishedSpans =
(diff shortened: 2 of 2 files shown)


opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
```diff
@@ -73,6 +71,37 @@ object TracingTest extends DefaultRunnableSpec {
...
+                  span.addEvent("In legacy code")
+                  Thread.sleep(10)
+                  span.addEvent("Finishing legacy code")
+                }.scopedSpan("Scoped")
+                  .span("Root")
+            spans  <- getFinishedSpans
+            root   = spans.find(_.getName == "Root")
+            scoped = spans.find(_.getName == "Scoped")
+            tags = spans.collect {
+              case span if span.getName == "Scoped" =>
```
runtologist, line 88: how about reusing `scoped` here?

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
```diff
@@ -73,6 +71,37 @@ object TracingTest extends DefaultRunnableSpec {
               )
             )
         },
+        testM("scopedSpan") {
+          for {
+            tracer <- ZIO.service[Tracer]
+            _ <- UIO.effectTotal {
+                  val span = tracer.getCurrentSpan
+                  span.addEvent("In legacy code")
+                  Thread.sleep(10)
```
runtologist, line 80: How about testing if the thread local is available here?

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
```diff
@@ -98,6 +126,24 @@ object Tracing {
       r   <- createChildOf(old, spanName, spanKind).use(finalizeSpanUsingEffect(effect, _, toErrorStatus))
     } yield r
 
+  /**
+   * Sets the current span to be the child of the current span with name
+   * 'spanName', introducing a thread-local scope during the execution allowing
+   * non-zio context propagation. Ends the span (and scope) when the effect
+   * finishes.
```
runtologist, line 133: The intended behaviour will break if `effect` for some reason switches the ExecutionContext or contains an interruptible region. Maybe that should be noted in the Scaladoc.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
```diff
@@ -98,6 +126,24 @@ object Tracing {
...
+   */
+  def scopedSpan[R, E, A](
+    spanName: String,
+    spanKind: Span.Kind,
+    toErrorStatus: PartialFunction[E, Status]
+  )(effect: ZIO[R, E, A]): ZIO[R with Tracing with Blocking, E, A] =
```
runtologist, line 139: Do we really need the blocking ExecutionContext here, or would an uninterruptible effect be sufficient?

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
easel: Good call. Will fix.

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
easel: Maybe? I'm still trying to determine if this is even a viable approach... I assume you're thinking along the lines of `assert(span != null) && assert(span.name == "Scoped")`?

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
easel: Good call. Interestingly enough, the java instrumentation package has advice for most implementations of execution context switching (akka, java util future, scala future, runnable, callable, etc). I've wondered if perhaps the better implementation of this would be some sort of zio-aware advice... Still trying to figure out how that all works.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
easel: Chances are you would know better than I! 

The way I was reading the docs, in order to guarantee staying on the same thread requires a .lock call, which requires an EC, and the blocking context was the easiest way to get access to one. I'll try dropping blocking and using `.uninterruptable` instead and see if the tests pass -- they do not pass with the `blocking` removed entirely.

easel:
Thanks for the quick review here @runtologist ! 

I want to bring up another issue here, which I'm also going to bring up with the java instrumentation folks on gitter. While this implementation appears to work both via the test and in a larger system (I get nicely connected Jaeger traces), in my larger system I'm getting spammed with a lot of `ERROR io.opentelemetry.auto.shaded.io.grpc.ThreadLocalContextStorage - Context was not attached when detaching`.

It seems to happen mostly when closing the scope and `Context.current() == Context.ROOT`, and it seems to only happen with spans originating in zio, vs spans that have zio in the middle of a stack initiated by an akka http call and instrumented automatically. So far I haven't been able to divine a conditional that I can use to successfully make it go away, it just kinda moves around. Any insight you might have would be appreciated.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
easel: I've switched to uninterruptible, which is still passing the tests. I will do some more testing outside and see how it works. Interestingly, most recently the tests pass without anything, but I don't think they create a sufficiently complex environment to cause interruptions consistently.

easel:
I was able to track down the cause of the `ERROR io.opentelemetry.auto.shaded.io.grpc.ThreadLocalContextStorage - Context was not attached when detaching` errors I've been seeing. The good news is that it appears as if the code in this PR works. The bad news is that the culprit appears to `ZIO.fromFuture`, which may need some instrumentation support to work properly. I'm still investigating that.

Wrapping slick calls, like this causes the error:
```
    Tracing.scopedSpan("SlickPostgres.runZ", Span.Kind.CLIENT, errorMapper)(
      for {
        result <- ZIO.fromFuture { implicit ec =>
          dbWithAccess.run(a)
        }
      } yield result
    )
```

But like this (using zio.blocking + Await.result) it works fine:
```
    blocking(
      Tracing.scopedSpan("SlickPostgres.runZ", Span.Kind.CLIENT, errorMapper)(
        Task.effect(Await.result(dbWithAccess.run(a), 10.seconds))
      )
    )
```

easel:
For reference, the full stack trace for the errors follows:

easel:
Unfortunately, switching the zio runtime to use a `ForkJoinPool` does not resolve the issue:

easel:
In interests of completeness,

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
easel: I've added an assertion that the current context != the root context, which seems to work pretty well in that if you swap `scopedSpan` for `span` the test blows up.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
easel: I've switched the implementation to match how the upstream project handles this, by wrapping the runnables. This eliminates the need for blocking or uninterruptible effects in favor of locking the executor with a wrapped copy of the same executor that wraps all the runnables it sees.

easel:
Ok. I've changed the approach to be more consistent with upstream, and added a test demonstrating the scope propagation works across `ZIO.fromFuture`. We're no longer dependent on blocking on uninterruptible effects, and generally should be able to propagate the scope across any normal code. I suppose there's a discussion to be had as to whether or not to make this the default behavior. It's probably what a user would generally expect, but it's going to introduce a bit of overhead so I'm not sure it's worth forcing it. 

I'm going to use this in anger a bit more, and am planning to connect with the otel guys tonight at their office hours, but assuming nothing horrible crops up between now and then I think this is ready to go pending final review.

runtologist:
Here is an idea that just came to my mind: We have all kinds of issues when the provided effect switches Threads inside ZIO. The main goal of this change is interop with legacy code. How about building something like an `OpenTelemetry.effect` and `OpenTelemetry.effectTotal`, which does the proper wrapping just for legacy code but does not accept a ZIO. That way breaking something at least becomes a lot more complicated. This approach could be implemented really easily and without additional overhead. What do you think?

easel:
That's an interesting idea. I experimented some with replacing `ZIO.fromFuture`, which would probably be the third "entrypoint" to have covered. I'm hoping to learn something about the expectations on the java side tonight that will inform this -- unfortunately the "lock to a span wrapping executor" approach works great for the trivial test case I added but so far does not work when wrapping slick calls.

easel:
Ok, so, maybe the third try is the charm! This patch is now basically an implementation of the suggestion in https://github.com/zio/zio-telemetry/pull/208#issuecomment-673674978. In discussion with the java auto-instrumentation folks, the key missing piece in the previous use of `tracer.withSpan` is that when mixing it with construction of Futures it has to be closed immediately after construction of the Future is complete.

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
```diff
@@ -73,6 +73,95 @@ object TracingTest extends DefaultRunnableSpec {
...
             )
         },
+        testM("scopedEffect") {
+          for {
+            tracer <- ZIO.service[Tracer]
+            _ <- Tracing.scopedEffect {
+                  val span = tracer.getCurrentSpan
+                  span.addEvent("In legacy code")
+                  if (Context.current() == Context.ROOT) throw new RuntimeException("Current context is root!")
+                  Thread.sleep(10)
```
runtologist, line 83: Why do we need to sleep here? I'm not sure how `Context` works, but might it make sense to try to manipulate it from some other thread concurrently? From the test I can see no reason, why it should change.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
```diff
@@ -98,6 +104,77 @@ object Tracing {
...
+   *
+   * CLoses the scope when the effect finishes
+   */
+  def scopedEffectFromFuture[R, A](make: ExecutionContext => scala.concurrent.Future[A]): ZIO[Tracing, Throwable, A] =
+    for {
+      currentSpan <- getCurrentSpan
+      tracer      <- getTracer
+      eff <- ZIO.fromFuture { implicit ec =>
+              // TODO: best practice is to use java try-with-resources. Is there
+              // something from ZIO or stdlib we can use that handles errors in close?
```
runtologist, line 168: How about `ZManaged` around the `scope`?

runtologist (commented):
This looks much nicer than before. :)

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
easel: The test is probably equally valid without the sleep, it seemed like a reasonable way to ensure we didn't optimize out the one liner. We could try to gin up something stressing the concurrency but I'm not sure it's worth the effort in this context.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
easel: I considered ZManaged, but I haven't been able to figure out how to use it without tying the lifetime to the zio runtime instead of the local block.  I think technically a "bracket" does what we need here but again, it's tightly tied to the ZIO and I haven't figured out the right way to unwind it.

I do have fairly good evidence that the current approach works so given this works I can take another look at optimizing it.

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
runtologist: OK. If you want to keep it, then please let's shorten it to a couple 100 ms. ;-)

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
runtologist: According to the JavaDoc, close cannot throw an `Exception`, so I guess the current code should be fine.

easel: I've pushed an update without the comments and tightened up the try-finally. Given close doesn't throw, as you say, I think keeping it simple and matching the upstream recommendation is probably better than trying to handle with with zmanaged or bracket and having to rely on the zio runtime to ensure everything runs on the correct thread.

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/TracingTest.scala
easel: I've gone ahead and removed it, I think it's probably unnecessary and obviously an unpleasant code smell.

easel:
I've rebased this against master and squashed the update commits. Barring further review comments, I think this is good to go and have flipped off draft.

runtologist (approved):

PR #211, opened 2020-08-14
Expose extension methods in the service

### LINKED ISSUE #170: Expose extension methods logic in the service
It would be nice if the functions exposed currently with an implicit class were also exposed in the service, with the `ZIO` as parameter. I have some use case where I don't want to have `OpenTracing` in the environment and I have an `OpenTracing.Service` object, but I'm forced to use `provideSomeLayer` to eliminate the dependency since the logic isn't exposed directly on the service.

mijicd:
Resolves #170 

### Summary

- Update service interface to expose only methods accepting zio.
- Rearranged methods alphabetically.
- Scraped some boilerplate (e.g. introduced `ZIO.service` calls).

### Questions

- Can we reduce visibility of `OpenTracing.managed`?
- Can we "hide" `error` and `finish` methods as they seem to be very much internal?
- Should we remove `tag`, `log` and `setBaggageItem` accessors that use `ZIO.unit` as default effect, or add the similar accessors for other methods?

### DIFF
--- a/opentracing-example/src/main/scala/zio/telemetry/opentracing/example/http/StatusesService.scala
+++ b/opentracing-example/src/main/scala/zio/telemetry/opentracing/example/http/StatusesService.scala
@@ -30,8 +30,7 @@ object StatusesService {
       case GET -> Root / "statuses" =>
         val zio =
           for {
-            env     <- ZIO.environment[OpenTracing]
-            _       <- env.get.root("/statuses")
+            _       <- ZIO.environment[OpenTracing].root("/statuses")
             _       <- OpenTracing.tag(Tags.SPAN_KIND.getKey, Tags.SPAN_KIND_CLIENT)
             _       <- OpenTracing.tag(Tags.HTTP_METHOD.getKey, GET.name)
             _       <- OpenTracing.setBaggageItem("proxy-baggage-item-key", "proxy-baggage-item-value")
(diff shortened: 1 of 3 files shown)


ghostdogpr:
I don't think it addresses what I needed? Unless I misunderstood 😄 

Basically I have this code:
```scala
zio.span(operation).provideSomeLayer[R](ZLayer.succeed(openTracing))
```
that I would like to change to:
```scala
openTracing.span(operation, zio)
```
So that I don't need to use `provideLayer` if I already have a service.

mijicd:
@ghostdogpr Let me see if I follow. I did exactly that for mentioned methods, and you need the other ones in the same form as well?

ghostdogpr:
Basically I'm interested in this logic which is in the extension method:
```scala
    def span(
      operation: String,
      tagError: Boolean = true,
      logError: Boolean = true
    ): ZIO[R with OpenTracing, E, A] =
      for {
        service <- getService
        old     <- getSpan(service)
        child   <- service.span(old, operation)
        r       <- span(service)(child, tagError, logError)
      } yield r
```
It's only available there, so you need the layer to use it. There is a `def span(service: OpenTracing.Service)` just next but it doesn't do the same thing (which is a bit dangerous by the way, since they have the same name). See what I mean?

mijicd:
@ghostdogpr I got it, that's what I enabled for `log`, `tag` and `setBaggageItem`, that's what I'm asking to confirm :). If so, I'll submit the changes for other methods in ~15.

ghostdogpr:
Ah okay, sorry 😆 I was focusing on the only one I need. This looks good then 🙏

mijicd:
In fact, I'm thinking to completely remove the old methods (without zio as param). Syntax class is what makes them useful, we can basically "unify" the interfaces.

mijicd:
@ghostdogpr @runtologist Updated the interface as suggested and reflected the changes in PR summary above.

ghostdogpr (approved):
LGTM but I have low expertise on this :)

runtologist (approved):
LGTM

Issue #292, opened 2021-01-28
Publish opencensus package


pettermachado:
I am trying to use zio-opencensus using the following dependency

```
"dev.zio" %% "zio-opencensus" % "0.7.0"
```
but get the following error when running `sbt update`

```
[error] sbt.librarymanagement.ResolveException: Error downloading dev.zio:zio-opencensus_2.13:0.7.0
[error]   Not found
[error]   Not found
[error]   not found: /home/petter/.ivy2/local/dev.zio/zio-opencensus_2.13/0.7.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/dev/zio/zio-opencensus_2.13/0.7.0/zio-opencensus_2.13-0.7.0.pom
...
```

It looks like zio-opencensus is not published at all in the `sbt ci-release` command, but I am not really sure where and how that is configured. Anyway, I would appreciate if zio-opencensus could get published.

mijicd:
Resolved in #294

PR #294, opened 2021-01-29
Add opencensus to root project

pettermachado:
Adding `opencensus` here made it work in `sbt publishLocal`. I do not have permissions to try `sbt ci-release` but :crossed_fingers: it will work for that case too.

See: https://github.com/zio/zio-telemetry/issues/292

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,7 @@ lazy val root =
   project
     .in(file("."))
     .settings(skip in publish := true)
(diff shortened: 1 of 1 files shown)


mijicd (approved):

mijicd:
@pettermachado Thank you for taking care of this. The snapshot will be published soon, and I'll cut the release as soon as everything's ok.

PR #319, opened 2021-02-21
Allow read/write access to current span

easel:
- Allows integration with spans provided by opentelemetry java api

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
@@ -31,7 +31,22 @@ object Tracing {
   private def createChildOf(parent: Span, spanName: String, spanKind: SpanKind): URManaged[Tracing, Span] =
     ZManaged.accessManaged[Tracing](_.get.createChildOf(parent, spanName, spanKind))
 
-  private def getCurrentSpan: URIO[Tracing, Span] = currentSpan.flatMap(_.get)
+  /**
+   * Get the current span stored in the zio-telemetry ref
+   * @return
+   */
+  def getCurrentSpan: URIO[Tracing, Span] = currentSpan.flatMap(_.get)
+
+  /**
+   * Get the current span stored in the zio-telemetry ref, allowing
+   * integration with externally created spans from native
+   * opentelemetry-api instrumentations
+   *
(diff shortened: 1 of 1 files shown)


duxet:
I was thinking about doing the same thing in my "Replace span with context" PR, in order to be able to log trace/span id:

```Scala
def wrapLogger[A](logger: Logger[A]): ZIO[Tracing, Nothing, Logger[A]] =
    logger.deriveM { logContext =>
      for {
        span <- Tracing.getCurrentSpan
      } yield annotateLogContext(span)(logContext)
    }

  private def annotateLogContext(span: Span)(logContext: LogContext) =
    logContext
      .annotate(traceId, span.getSpanContext.getTraceId)
      .annotate(spanId, span.getSpanContext.getSpanId)
```

easel:
Yeah, it's a nice idea to keep the internal span private but once you start interacting with the rest of the world not having access to it becomes problematic.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
```diff
@@ -31,7 +31,22 @@ object Tracing {
...
+
+  /**
+   * Get the current span stored in the zio-telemetry ref, allowing
+   * integration with externally created spans from native
+   * opentelemetry-api instrumentations
+   *
+   * @param span The span to set as current
+   * @return
+   */
+  def setCurrentSpan(span: Span): URIO[Tracing, Unit] =
```
runtologist, line 48: Would it make sense to stop the previous span or to restore it later?

easel: I'm not sure. The use case I've had for this is integrating with spans created via the play and akka http instrumentation, so we'd only be replacing the default span. There might be a better way to "re-initialize" the span via the zio runtime, too, but I didn't have much luck with it.

mijicd:
@runtologist @easel what's the state of this pull request?

runtologist:
> @runtologist @easel what's the state of this pull request?

I do not have a strong opinion, it can be merged I guess.

easel:
Closing this out. Need to rethink if it's necessary if https://github.com/zio/zio-telemetry/pull/580 works properly.

PR #369, opened 2021-05-19
Make getCurrentContext/Span public

duxet:
I don't think there is any reason to limit access to current context/span outside of zio-telemetry.


duxet:
@mijicd @runtologist ping :D

runtologist:
Initially we had planned to hide as many internals as possible with the idea of forming a common API that abstracts from opentelemetry and opentracing. As the implementations have diverged anyway, This change works for me. @mijicd wdyt?

mijicd:
@runtologist I agree as well.

mijicd (approved):

PR #418, opened 2021-08-19
Have spanFrom work like root when the tracer's extract returns null

balajirrao:
When the tracer's extract method returned null, spanFrom was a no-op. Fix it
to work like root in this case.


balajirrao:
if the new behaviour of `spanFrom` is acceptable, I'll add a test for it in this PR.

runtologist (commented):
Thanks for this contribution. Sounds reasonable, please go ahead.

balajirrao:
@runtologist, I've added the tests. Happy to hear your feedback!

balajirrao:
@runtologist, gentle reminder to review this PR 😄

opentracing/src/main/scala/zio/telemetry/opentracing/OpenTracing.scala
```diff
@@ -103,10 +103,9 @@ object OpenTracing {
...
-        .fold(_ => None, Option.apply)
-        .flatMap {
-          case None          => zio
-          case Some(spanCtx) =>
+        .foldM(
+          _ => zio,
+          spanCtx =>
             for {
               current <- service.currentSpan.get
               span    <- UIO(service.tracer.buildSpan(operation).asChildOf(spanCtx).start())
```
nsauro, line 111: Please pardon my lurking, but I've been playing around with this library, and tracing in general. I too ran into this same problem, so this PR interested me. 

According to the documentation: https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/Tracer.java#L138-L152 if you pass in a `null` `SpanContext` or `Span`, it is a no op.

Do you think passing in a create function into the `spanFrom` would help?

balajirrao: The `childOf` would be a no-op, which is what we indeed want. Without the PR, even `buildSpan` wouldn't be called if `spanCtx` was null. Does this help or did I not understand your question at all? 😄

runtologist (approved):
LGTM. Thank you and sorry this too so long.

opentracing/src/main/scala/zio/telemetry/opentracing/OpenTracing.scala
nsauro: Ah, looking at this with fresher eyes, I see the intention.  This was me just getting tripped up on ZIO code which I'm still learning.  Looks much better now!

balajirrao:
@runtologist, no problem, thank you! I would have thought the next step would be to merge it - but I can't, due to lack of write access.

Issue #377, opened 2021-05-28
Scala 3


francisdb:
Is a Scala 3 release planned?
What are the roadblocks?

runtologist:
A Scala 3 release would be great. I guess the only change necessary is extending the build configuration. A PR would be highly appreciated.

mvelimir:
I will work on this.

mvelimir:
There is a type mismatch error when compiling the opentracing tests. Opentelemetry tests pass just fine. Also, some of the libraries used in the example applications don't have Scala 3 support, such as zio-config-magnolia, zio-magic and zio-json from sttp.client3.

francisdb:
@mvelimir found related scala 3 tickets on those projects?

PR #430, opened 2021-09-13
Crossbuild for Scala3

KadekM:
Would this be sufficient for Scala3 version?
Closes https://github.com/zio/zio-telemetry/issues/377, Closes #432

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,12 +36,13 @@ Global / testFrameworks       := Seq(new TestFramework("zio.test.sbt.ZTestFramew
 
 addCommandAlias("fmt", "all scalafmtSbt scalafmt test:scalafmt")
 addCommandAlias("check", "all scalafmtSbtCheck scalafmtCheck test:scalafmtCheck")
+addCommandAlias("compileExamples", "opentracingExample/compile;opentelemetryExample/compile")
 
 lazy val root =
   project
     .in(file("."))
     .settings(publish / skip := true)
-    .aggregate(opentracing, opentelemetry, opencensus, opentracingExample, opentelemetryExample)
+    .aggregate(opentracing, opentelemetry, opencensus)
 
 lazy val opentracing        =
   project
@@ -65,6 +66,7 @@ lazy val opentracingExample =
     .in(file("opentracing-example"))
     .settings(stdSettings("opentracing-example"))
     .settings(publish / skip := true)
+    .settings(onlyWithScala2)
     .settings(libraryDependencies := Dependencies.opentracingExample)
     .dependsOn(opentracing)
 
@@ -73,6 +75,7 @@ lazy val opentelemetryExample =
     .in(file("opentelemetry-example"))
     .settings(stdSettings("opentelemetry-example"))
     .settings(publish / skip := true)
+    .settings(onlyWithScala2)
     .settings(libraryDependencies := Dependencies.opentelemetryExample)
     .dependsOn(opentelemetry)
 
--- a/opentracing/src/test/scala/zio/telemetry/opentracing/OpenTracingTest.scala
+++ b/opentracing/src/test/scala/zio/telemetry/opentracing/OpenTracingTest.scala
@@ -20,7 +20,7 @@ object OpenTracingTest extends DefaultRunnableSpec {
     ZLayer.fromEffect(UIO(new MockTracer))
 
   val testService: URLayer[HasMockTracer with Clock, OpenTracing] =
-    ZLayer.fromServiceManaged(tracer => OpenTracing.managed(tracer, "ROOT"))
+    ZLayer.fromServiceManaged((tracer: MockTracer) => OpenTracing.managed(tracer, "ROOT"))
 
   val customLayer = mockTracer ++ ((mockTracer ++ Clock.any) >>> testService)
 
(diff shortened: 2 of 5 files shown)


opentracing/src/test/scala/zio/telemetry/opentracing/OpenTracingTest.scala
```diff
@@ -20,7 +20,7 @@ object OpenTracingTest extends DefaultRunnableSpec {
     ZLayer.fromEffect(UIO(new MockTracer))
 
   val testService: URLayer[HasMockTracer with Clock, OpenTracing] =
-    ZLayer.fromServiceManaged(tracer => OpenTracing.managed(tracer, "ROOT"))
+    ZLayer.fromServiceManaged((tracer: MockTracer) => OpenTracing.managed(tracer, "ROOT"))
```
KadekM, line 23: gets rid of compiler error on Scala3

KadekM:
How about skipping those 3.0.2 tests until the dependant libs are published for Scala3? (zio-config, zio-magic, zio-json, sttp...)

project/BuildHelper.scala
```diff
@@ -12,6 +12,10 @@ object BuildHelper {
     incOptions ~= (_.withLogRecompileOnMacro(false))
   )
 
+  val withScala3 = Seq(
+    crossScalaVersions       := Seq(Scala212, Scala213, Scala3),
+  )
+
```
mijicd, line 18: Instead of this, could you please bring zio-telemetry to the same standard as other projects and load scala versions from actions file? Basically you need to do the following:

- [add snakeyaml plugin](https://github.com/zio/zio-redis/pull/385/files#diff-8fd74420ca55aa91d4c53775d2a55a999c405b5a4fafeaf73b7f0ef83e1f3b0cR11)
- [define versions map](https://github.com/zio/zio-redis/pull/385/files#diff-dd3701b993fcc974a9a6d017159aac00d4b41bcd05ad319815c6001311c669d5R8)

We introduced this change after zio-telemetry was ported to Github Actions and we didn't update it :man_facepalming: .

project/BuildHelper.scala
```diff
@@ -74,6 +81,9 @@ object BuildHelper {
           "242"
         ) ++ std2xOptions ++ optimizerOptions(optimize)
 
+      case Some((3, _)) =>
```
mijicd, line 84: Please stay [aligned with ZIO](https://github.com/zio/zio/blob/master/project/BuildHelper.scala#L139) as much as possible.

mijicd (commented):
It looks like the example projects are causing all build problems. If that's the case, I don't mind disabling Scala 3 build for them, but please write up the missing dependencies as we might be able to "push" the maintainers for some of them.

project/BuildHelper.scala
KadekM: We already have the project map - https://github.com/zio/zio-telemetry/blob/master/project/BuildHelper.scala#L15-L26, I did it so that we don't build example projects for Scala3

mijicd: Oh sorry, good point. I would just swap the two - `stdOptions` should include Scala 3, and we should be explicit about missing Scala 3 in examples.

KadekM:
@mijicd the problem deps are:
```
zio-config-magnolia
zio-json
com.softwaremill.sttp.client3/zio-json
zio-magic
```

project/BuildHelper.scala
KadekM: @mijicd good point, I swapped it. Also I bumped scala version to 2.12.15

mijicd:
@KadekM this looks good to me, please rebase & format, and it's good to go!

KadekM:
> @KadekM this looks good to me, please rebase & format, and it's good to go!

will do - do I need to disable the tests for Scala3 or can we do merge/release ignoring the pipeline failures for 3.0.2 test ?

KadekM:
just linking these two https://github.com/zio/zio-config/issues/506 https://github.com/zio/zio-json/issues/434 if they arrive sooner we can swap

mijicd:
It seems that those tests are failing because of example dependencies. That said, maybe it's simpler if we just don't aggregate examples, and compile and check them separately for 2.x.

KadekM:
I've disabled the aggregation of them in root

build.sbt
```diff
@@ -41,7 +41,7 @@ lazy val root =
   project
     .in(file("."))
     .settings(publish / skip := true)
-    .aggregate(opentracing, opentelemetry, opencensus, opentracingExample, opentelemetryExample)
+    .aggregate(opentracing, opentelemetry, opencensus)
```
mijicd, line 44: Note that this will exclude them completely. Make sure to create compilation steps for these projects and run them separately, e.g.:

```
// in build.sbt
addCommandAlias("compileExamples", "opentracingExample/compile;opentelemetryExample/compile")

// in ci config
name: Compile examples
if: ${{ startsWith(matrix.scala, '2.') }}
run: ./sbt ++${{ matrix.scala }}! compileExamples
```

mijicd: @KadekM :arrow_up: :smile:

KadekM: thanks for writing the YAML part, I was afraid to touch it :D

mijicd (approved):
Thank you for taking care of this!

KadekM:
Thank you for taking care of me taking care of this!

runtologist:
Thanks a lot!

PR #437, opened 2021-09-21
feat: add spanFromUnsafe and spanUnsafe

frekw:
Hi!

I needed this when adding interop with zio-http.

```scala
object Traced {
  def apply[R, E](app: HttpApp[R, E]): HttpApp[R with Tracing, E] =
    Http
      .fromEffectFunction[Request] { req =>
        ZIO
          .succeed(app)
          .spanFrom(propagator, req.headers, getter, req.endpoint._2.path.toString, SpanKind.SERVER)
      }
      .flatten
}
```

In this example, the `app` will continue to be executed after this effect has returned, which means every subsequent span created will end up without a root.

This allows one to manually control the spans, which means we can implement a general tracing wrapper like so:

```scala
  def apply[R, E](app: HttpApp[R, E]): HttpApp[R with Tracing, E] =
    Http
      .fromEffectFunction[Request] { req =>
        for {
          trace             <- startSpan(req)
          (currentSpan, pop) = trace
          res                = app.foldM(
                                 ee => Http.fromEffect(setErrorStatus(currentSpan, ee) *> ZIO.fail(ee).ensuring(pop)),
                                 res => Http.fromEffect(ZIO.succeed(res).ensuring(pop)),
                                 Http.fromEffect(pop).as(Http.empty).flatten
                               )
        } yield res
      }
      .flatten

  private def startSpan(req: Request) =
    Tracing.unsafeSpanFrom(propagator, req.headers, getter, req.endpoint._2.path.toString, SpanKind.SERVER)

  private def setErrorStatus[E](
    span: Span,
    err: E
  ): UIO[Span] =
    UIO(span.setStatus(StatusCode.ERROR, err.toString))
```

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
@@ -17,11 +17,13 @@ object Tracing {
     private[opentelemetry] val currentContext: FiberRef[Context]
     private[opentelemetry] def createRoot(spanName: String, spanKind: SpanKind): UManaged[Context]
     private[opentelemetry] def createChildOf(parent: Context, spanName: String, spanKind: SpanKind): UManaged[Context]
+    private[opentelemetry] def createChildOfUnsafe(parent: Context, spanName: String, spanKind: SpanKind): UIO[Context]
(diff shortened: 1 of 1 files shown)


runtologist (approved):
LGTM, thank you.

frekw:
🙏

PR #320, opened 2021-02-21
Create child spans from external with inSpan

easel:

PR #441, opened 2021-10-07
Adds Tracing.inSpan, a way to give zio-telemetry an externally created span

cursedquail:
This builds on the work in #320, adding tests & some documentation


NLincoln:
@runtologist ah, I forgot to run linting locally. Should be gtg now? Also npm is operational again so the docs bit should be coolio

Issue #338, opened 2021-03-29
OpenTelemetry: Baggage API


duxet:
It would be nice to have some functions to manipulate the [OpenTelemetry Baggage](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md).

PR #462, opened 2021-11-10
Add baggage API #338

duxet:
I've added two new functions to get and set the baggage entries in the current context.


opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
```diff
@@ -305,6 +307,23 @@ object Tracing {
     getCurrentSpan.map(_.setAttribute(AttributeKey.doubleArrayKey(name), v))
   }
 
+  /**
+   * Sets a baggage entry in the current context
+   */
+  def setBaggage(name: String, value: String): URIO[Tracing, Context] =
+    for {
+      contextRef <- currentContext
+      context    <- contextRef.getAndUpdate(context =>
```
runtologist, line 316: Do we want to return the context before or after the change here?

mijicd (approved):

PR #485, opened 2022-01-09
chore: upgrade to zio 2.0.0-rc1

vladimir-lu:
This can't be merged because the examples don't compile (due to missing ZIO 2 version of zio-http). But perhaps someone can publish it as a snapshot version?

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/package.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/package.scala
@@ -1,7 +1,5 @@
 package zio.telemetry
 
-import zio.Has
-
 package object opentelemetry {
-  type Tracing = Has[Tracing.Service]
+  type Tracing = Tracing.Service
 }
--- a/opentracing/src/main/scala/zio/telemetry/opentracing/package.scala
+++ b/opentracing/src/main/scala/zio/telemetry/opentracing/package.scala
@@ -1,10 +1,10 @@
 package zio.telemetry
 
 import io.opentracing.propagation.Format
-import zio.{ Has, ZIO }
+import zio.ZIO
 
 package object opentracing {
-  type OpenTracing = Has[OpenTracing.Service]
+  type OpenTracing = OpenTracing.Service
 
   implicit final class OpenTracingZioOps[-R, +E, +A](val zio: ZIO[R, E, A]) extends AnyVal {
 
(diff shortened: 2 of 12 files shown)


runtologist:
Hey, thank you for this contribution, I'm looking forward to a ZIO 2 version! 

We should still be able to do bugfix releases for 1.0 afterwards. I guess it makes sense to create a series 1.x branch, that stays on ZIO 1 and continue with ZIO 2 in master. @mijicd Do you know if it's possible to configure renovate for two branches? on 1.x we need to set it up to use ZIO 1 dependencies.

mijicd:
@runtologist Unfortuntely, I think we'll have to do the same trick here.

swoogles:
ZioHttp and Sttp both have RC1 releases now, so you should be clear to upgrade @vladimir-lu  :)

vladimir-lu:
Excellent, I've updated the examples. Please have a look because I have a feeling the way I've written it is not idiomatic.

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/http/package.scala
```diff
@@ -1,7 +1,5 @@
 package zio.telemetry.opentelemetry.example
 
-import zio.Has
-
 package object http {
-  type Client = Has[Client.Service]
+  type Client = Client.Service
```
mijicd, line 4: I'd remove this alias completely.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/package.scala
```diff
@@ -1,7 +1,5 @@
 package zio.telemetry
 
-import zio.Has
-
 package object opentelemetry {
-  type Tracing = Has[Tracing.Service]
+  type Tracing = Tracing.Service
```
mijicd, line 4: Same here.

opentracing/src/main/scala/zio/telemetry/opentracing/package.scala
```diff
@@ -1,10 +1,10 @@
 package zio.telemetry
 
 import io.opentracing.propagation.Format
-import zio.{ Has, ZIO }
+import zio.ZIO
 
 package object opentracing {
-  type OpenTracing = Has[OpenTracing.Service]
+  type OpenTracing = OpenTracing.Service
```
mijicd, line 7: Ditto.

vladimir-lu:
Removed the type aliases now

PR #499, opened 2022-02-15
Set error attribute when encountering an error with a span

cursedquail:
👋 we've been using zio-open-telemetry with DataDog for a bit. One weird thing we've been noticing is that our error spans don't show up as having errored in the DataDog UI. 
I think this is because DD expects

```scala
span.setAttribute("error", true)
```

When you encounter an error, which zio-open-telemetry doesn't do currently. Is there interest in adding those to zio-telemetry, probably in here? https://github.com/zio/zio-telemetry/blob/7e6e29c60ad4d9a72c03b175f598fb51f00ea693/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala#L56

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/Tracing.scala
@@ -53,7 +53,11 @@ object Tracing {
     toErrorStatus: PartialFunction[E, StatusCode]
   ): UIO[Span] = {
     val errorStatus: StatusCode = cause.failureOption.flatMap(toErrorStatus.lift).getOrElse(StatusCode.UNSET)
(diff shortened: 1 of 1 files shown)


runtologist:
Yes, this looks good. Would you mind adding a test and un-drafting?

n3ziniuka5:
I think the actual problem is on line 55:
```scala
val errorStatus: StatusCode = cause.failureOption.flatMap(toErrorStatus.lift).getOrElse(StatusCode.UNSET)
```

If you change `StatusCode.UNSET` to `StatusCode.ERROR` then you don't need any additional attributes and the errored traces appear as expected in Datadog.

cursedquail:
Closing this PR - I no longer work with ZIO (changed jobs) so I won't be able to put the time in to finish it - sorry for leaving it open for so long :disappointed:

PR #516, opened 2022-04-05
ZIO-2.0.0-RC4

vladimir-lu:


runtologist:
Thanks a lot for this upgrade @vladimir-lu. `sbt compileExamples` is still failing. Would you mind taking a look at that?

vladimir-lu:
Should be good now. I think `ZIO.scoped` acts a little bit differently in type inference compared to `ZManaged` but overall it does seem to be a good simplification.

Would you be able to cut a release after this is merged?

runtologist (approved):
LGTM

runtologist:
> Should be good now. I think `ZIO.scoped` acts a little bit differently in type inference compared to `ZManaged` but overall it does seem to be a good simplification.

Yes, I found in a few places it needs type annotations, where ZManaged didn't.

Thanks again for the upgrade. 

> Would you be able to cut a release after this is merged?

I'll do.

PR #521, opened 2022-04-17
ZIO-2.0.0-RC5

vladimir-lu:
The zio-http dependency has not released yet for RC5 (even though the version bump has been committed) - but this doesn't break example compilation, just runtime.


mijicd:
Please revert this change and synchronize with zio-http team on their plans to release an upgrade.

vladimir-lu:
Ok, apologies about that - reverted the changes and new PR for the same once zio-http updates: https://github.com/zio/zio-telemetry/pull/522

PR #548, opened 2022-07-15
chore(deps): Bump dependency versions, update examples with more moderns versions

vladimir-lu:
Also:

* Bump sttp to 3.7.0 (which is actually required for ZIO 2.0 support - only showed up once I ran proxy server example)
* Add the service.name resource so that the opentelemetry example looks better in jaeger UI
* Moved example versions to own section so it's easier to see what is a core dependency when looking at at it

I know there's sbt 1.7 available now but it changes the cross-compilation handling so perhaps it might need more than a cursory look.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 **/.DS_Store
 **/metals.sbt
 target
+out/
 .bloop
 .metals
 .idea
@@ -11,4 +12,3 @@ website/build
 website/i18n
 website/static/api
 website/yarn.lock
-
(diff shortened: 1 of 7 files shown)


francisdb (approved):

francisdb:
Looking forward to a ZIO 2 release as we currently have

```
Exception in thread "zio-fiber-" java.lang.NoSuchMethodError: 'zio.ZIO zio.ZIO$.succeed(scala.Function0, java.lang.Object)'
	at zio.telemetry.opentelemetry.Tracing$Live$1.createRoot$$anonfun$2$$anonfun$2(Tracing.scala:10)

```

mijicd:
@vladimir-lu There are **multiple** pull requests made by steward which I am resolving and merging to `master` as we speak. By merging this pull request, `master` and version 1 is left in a weird state. Please don't do this.

PR #567, opened 2022-08-21
Merge series/2.x into master

grouzen:
The release version will be `2.0.x` since @vladimir-lu has recently added a `2.0.0` tag.


mijicd:
I would not do this. Releases are already enabled from series/2.x branch. What we could consider doing is renaming master, and making series/2.x default (the way how ZIO itself did it).

grouzen:
Hey @mijicd! Thanks for the help with this. I'll try to rename `master` to `series/1.x` as it is done in `zio` repo.

grouzen:
Basically I've come up with this migration plan:
- [x] create branch protection rules for `series/2.x` the same as for `master`
- [ ] check already created PRs for `master`
	- [ ] maybe it is worth to merge them first before renaming
- [ ] make `series/2.x` default
- [ ] rename `master`
- [ ] update branch protection for `series/1.x`

Issue #561, opened 2022-08-06
Spans Created from zio-telemetry and auto instrumentation are not connected


grogdotcom:
My company uses mostly a Java stack and uses opentelemetry auto instrumentation for the majority of its services however I work on a ZIO service. We would also like to leverage auto instrumentation. We are able to get some connected spans through auto instrumentation but we have noticed when the root span is created with zio-telemetry and any downstream span is created with auto instrumentation, the spans will not be connected. 

I see #207 solves this issue when the auto instrumented spans are from non-zio code but when the auto instrumented spans are from ZIO code this solution no longer works.  As an example, lets say I have a non-zio database library that is wrapped in a ZIO library, so all my calls to the database are ZIOs. If I create a new span and call into this ZIO database wrapper library, the spans in jaeger will be separate. 

What is interesting however is that when my service receives a grpc call that downstream calls a ZIO wrapped database library, all of the auto instrumented spans are connected. However in this case zio-telemetry is never used. 

My guess on what is going on is that the active span set in the ZIO environment does not translate to the underlying Context for auto instrumentation.

dmytr:
Hey @grogdotcom,

My [PR](https://github.com/zio/zio-telemetry/pull/580) adds an experimental support for context propagation. It's merged and there is a snapshot version available. AFAIK a release is planned soon as well.

Could you let me know about your experience, if you decide to take it for a spin, please?

jeejeeone:
Issue still persists. Here is an example repo highlighting the issue https://github.com/jeejeeone/zio-telemetry-poc2 . After adding opentelemetry agent traces defined in zio code are separate and autoinstrumented traces are also separate. Sadly I don't have time to make the example more easily runnable but you get the gist.

dmytr:
Hi @jeejeeone,

I think the problem is that in this example a new Jaeger tracer is created with `JaegerTracer.live`:

```scala
    Server.serve(app).provide(Server.default, Tracing.propagating, JaegerTracer.live)
```
(Link: https://github.com/jeejeeone/zio-telemetry-poc2/blob/main/src/main/scala/com/example/Main.scala#L62)

I've tried using the  tracer provided by the automatic instrumentation and all traces seem to be correctly connected. To do this:

```scala
    Server.serve(app).provide(Server.default, Tracing.propagating, ZLayer.succeed(GlobalOpenTelemetry.getTracer("test")))
```

[...]

PR #580, opened 2022-09-06
Propagate OTEL context between ZIO and non-ZIO code

dmytr:
### Motivation

At the moment OTEL context is not propagated between ZIO and non-ZIO code. Of course, there are functions from `scopedEffect` family provided by `Tracing`, but they expect an unsafe size-effecting code as their argument.

In case if you need to propagate context into libraries that wrap non-ZIO code into ZIO, but don't use tracing - there is no guarantee that underlying non-ZIO code will have access to the correct context.

And in cases when non-ZIO code invokes ZIO code (for example in `zhttp`, which starts fibers from Netty code) context is not (correctly) propagated as well.

This change makes it possible to use automatic instrumentation provided by [OpenTelemetry JVM agent](https://opentelemetry.io/docs/instrumentation/java/automatic/).

Related: https://github.com/zio/zio-telemetry/issues/561

### How this works

This solution uses a special ZIO supervisor that takes a snapshot of the current OpenTelemetry context before a new ZIO fiber is started and uses it as a parent context in the scope of that fiber.

Each time a fiber is suspended, supervisor takes a fresh snapshot of the current context and then restores the root context. When fiber is resumed, it restores the last context snapshot for that fiber.

When fiber is completed, supervisor restores the root context.

Probably, this is not the most correct or efficient implementation, but it does seem to work.

There is an end-to-end setup that uses this kind of supervisor here: https://github.com/dmytr/zio-otel-instrumentation.

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/PropagatingSupervisor.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/PropagatingSupervisor.scala
@@ -1,0 +1,45 @@
+package zio.telemetry.opentelemetry
+
+import io.opentelemetry.api.trace.Span
+import io.opentelemetry.context.Context
+import zio._
+
+import java.util.concurrent.ConcurrentHashMap
+import scala.annotation.nowarn
+
+@nowarn("msg=discarded non-Unit value")
+final class PropagatingSupervisor extends Supervisor[Unit] {
+
+  private val storage = new ConcurrentHashMap[FiberId, Span]()
+
+  override def value(implicit trace: Trace): UIO[Unit] = ZIO.unit
+
+  override def onStart[R, E, A](
+    environment: ZEnvironment[R],
+    effect: ZIO[R, E, A],
+    parent: Option[Fiber.Runtime[Any, Any]],
+    fiber: Fiber.Runtime[E, A]
+  )(implicit unsafe: Unsafe): Unit = {
+    val span = Span.current()
+    if (span != null) storage.put(fiber.id, span)
+    else storage.put(fiber.id, Span.fromContext(Context.root()))
+  }
+
+  override def onEnd[R, E, A](value: Exit[E, A], fiber: Fiber.Runtime[E, A])(implicit unsafe: Unsafe): Unit = {
+    storage.remove(fiber.id)
+    Context.root().makeCurrent()
+  }
+
+  override def onSuspend[E, A](fiber: Fiber.Runtime[E, A])(implicit unsafe: Unsafe): Unit = {
+    val span = Span.current()
+    if (span != null) storage.put(fiber.id, span)
+    else storage.put(fiber.id, Span.fromContext(Context.root()))
+    Context.root().makeCurrent()
+  }
+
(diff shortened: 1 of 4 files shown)


opentelemetry/src/main/scala/zio/telemetry/opentelemetry/PropagatingSupervisor.scala
```diff
@@ -0,0 +1,45 @@
+package zio.telemetry.opentelemetry
+
+import io.opentelemetry.api.trace.Span
+import io.opentelemetry.context.Context
+import zio._
+
+import java.util.concurrent.ConcurrentHashMap
```
grouzen, line 7: I guess it can be replaced with https://zio.dev/reference/sync/concurrentmap. As a consequence, you will have `Supervisor[UIO[Unit]]` and `@nowarn` annotation might be omitted.

dmytr: I'm afraid we'll have to stick to the regular `ConcurrentHashMap` because all callbacks in `Supervisor` are [unsafe and return `Unit`](https://github.com/zio/zio/blob/a9c924d4d346c3561e298546236e15b29d3b72c4/core/shared/src/main/scala/zio/Supervisor.scala#L53-L66). Unsafe side effects have to be executed within those callbacks for context to be properly propagated into non-ZIO code.

dmytr: `@nowarn` basically tells us that we are throwing away `Scope` that we get when we call `Context.makeCurrent()`. Ideally we shouldn't do this. I saw that other tracing libraries (like Datadog's tracing SDK) use stacks of scopes. So potentially we could change `storage` to be of type `ConcurrentHashMap[FiberId, Stack[Scope]]`. But this is kind of involved and apparently this all works even when we do throw scopes away.

grouzen:
@dmytr Thanks for your contribution! Sorry for the delay, I honestly need some time to process this PR.

grouzen:
Thanks again, @dmytr, for your contribution! It looks good to me.

grouzen (approved):

easel (approved):
Very nice contribution. Given the interface surface area change is very small, I'd advocate for merging and releasing soon to make it easier to get more feedback. In general, I think the approach is the right one -- configure the zio runtime appropriately and context propagates as expected into and out-of the zio runtime.

grouzen (approved):

easel (approved):

vladimir-lu (approved):

grouzen:
@dmytr Hey! Can we merge it?

dmytr:
@grouzen yes, please. I don't have access to do this.

dmytr:
I've updated my test harness to use the published snapshot version: https://github.com/dmytr/zio-otel-instrumentation/pull/2
It works as expected in that specific setup :partying_face:

grouzen:
I'll cut a release ASAP.
Update: 
https://github.com/zio/zio-telemetry/releases/tag/v2.0.3

PR #583, opened 2022-09-12
Use Module Pattern 2.0 in the OpenTelemetry example

mvelimir:


mvelimir:
@dmytr Does this look good to you?

dmytr:
Hey @mvelimir, looks good to me. But I'm not a maintainer :sweat_smile: 

I'd only suggest to take a look on documentation and update it as well, if needed. So that examples in docs and example project are in sync. But it could be that examples in docs will be way more concise if they keep using accessors :man_shrugging:

grouzen:
Hey @mvelimir! Thanks for the PR. For me, it looks good. I just must say that to make it fully idiomatic, we should eliminate the type parameter `Tracing` in `BackedApp.routes`. But we cannot do this right away because  `response.spanFrom`  syntax requires it. I've already created an issue for this https://github.com/zio/zio-telemetry/issues/585. I guess it is okay to merge this PR now and do the rest later on.

grouzen (dismissed):

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/http/BackendApp.scala
```diff
@@ -4,18 +4,33 @@ import io.opentelemetry.api.trace.SpanKind
...
+import BackendApp._
 
 import java.lang
 import scala.jdk.CollectionConverters._
 
+final case class BackendApp(tracing: Tracing) {
+
+  val routes: HttpApp[Any, Throwable] =
+    Http.collectZIO { case request @ Method.GET -> !! / "status" =>
+      tracing.spanFrom(propagator, request.headers, getter, "/status", SpanKind.SERVER, Map.empty[Throwable, Nothing]) {
```
dmytr, line 20: Hm, requirement to add this `Map.empty[Throwable, Nothing]` doesn't seem to be that convenient for clients. Maybe we should add the default value of `Map.empty` to all `toErrorStatus` parameters in `Tracing`? :thinking:

grouzen: I'm going to find a better encoding for syntax, so this will be changed to `response.spanFrom` where `toErrorStatus` would have a default value as well as we have for all syntax methods. But, of course, we might want to add a default value for the `Tracing` methods. Hope it makes sense.

grouzen: Right after I wrote the previous comment, I realized that it makes sense to avoid having syntax at all since using the module pattern 2.0 approach leads us to have two options on how to call the methods:
- `tracing.spanFrom(...)(effect)`
- `effect.spanFrom(...)(tracing)`
which doesn't make any difference.

grouzen: I'll add default values for this and other methods in PR that I'm working on right now https://github.com/zio/zio-telemetry/pull/586, so this may remain as is for now.

grouzen (approved):

grouzen:
Sorry for saying this, but I think it would be better to cancel this PR since I've done this along with `opentracing-example` already in my [PR](https://github.com/zio/zio-telemetry/pull/586/files). It is better to close because I've made many changes to the modules and unified APIs of modules and, consequently, examples.

PR #591, opened 2022-10-18
Use module pattern 2.0 for OpenTelemetry

Doikor:
It was reverted to the old style by accident (I think?) in #580

Mainly just reverting back to previous version and moving `Live` outside of `scoped` so we can use the same one in `scopedPropagating`.


Doikor:
meh ignore this. It is kind of partially but not really implemented. Just kind of looked like it before the propagation change.

Issue #558, opened 2022-07-30
Integrate into ZIO Logging & Metrics


jdegoes:
The goal of ZIO Telemetry should be to seamlessly add OpenTelemetry support via ZIO's own built-in logging and metrics, adding integration points as necessary and useful.

TODO

 - [x] How to map ZIO logging + annotations + other context to OpenTelemetry traces
 - [x] How to map ZIO logging + annotations + other context to OpenTelemetry logs
 - [x] How to map ZIO metrics to OpenTelemetry metrics
 - [x] How to map ZIO context + annotations to OpenTelemetry baggage

easel:
These goals sound incredible. What's less clear to me is how complete the coverage is in native ZIO capabilities. I know ZIO has some "tracing" capabilities but am much less clear what they are and how well they map to the equivalent Otel concepts. @jdegoes any suggestions as to how we can get starting mapping these capabilities?

grouzen:
I'm integrating Baggage with ZIO log annotations in this PR: https://github.com/zio/zio-telemetry/pull/662

omidb:
is there any update on this?

grouzen:
Hey! I'm still working on adding metrics support in general.  After this, I believe mapping ZIO metrics to OTEL metrics should be possible. 

The only part that was done long ago is mapping ZIO annotations to OpenTelemetry baggage. It is not reflected in docs, though, but you can check out the source code at https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/baggage/Baggage.scala and find some examples in https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry/src/test/scala/zio/telemetry/opentelemetry/baggage/BaggageTest.scala#L89.

I was researching how to map ZIO logging and I don't think it is doable since there is no way to extract the original log message. I hope I am wrong about it.
Mapping annotations to traces and logs is also possible.

omidb:
@grouzen thanks for the update. I'm trying to understand how does this work with something like `myzio.logError()`. Should I `tapError` and then create a Baggage? something like this: `.tapError(e => baggage.set("some", e.getMessage))`

grouzen:
@omidb Yes, your solution is fine. As I said in the message above, there is no integration with ZIO logging at the moment and I don't think it is possible to make it happen due to technical reasons.

grouzen:
[...]

PR #662, opened 2023-03-02
Opentelemetry zio integrate log annotations to baggage

grouzen:
This PR solves one of the subtasks of https://github.com/zio/zio-telemetry/issues/558. 

The solution is to propagate ZIO log annotations as OTEL Baggage values automatically. For this, the end user should use  `Baggage.logAnnotated` or `Baggage.live(logAnnotated = true)` layer constructor.

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/baggage/Baggage.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/baggage/Baggage.scala
@@ -114,39 +114,44 @@ trait Baggage { self =>
 
 object Baggage {
 
-  def live: URLayer[ContextStorage, Baggage] =
+  def live(logAnnotated: Boolean = false): URLayer[ContextStorage, Baggage] =
     ZLayer {
       for {
         ctxStorage <- ZIO.service[ContextStorage]
       } yield new Baggage { self =>
         override def getCurrentBaggage(implicit trace: Trace): UIO[Baggaje] =
-          getCurrentContext.map(Baggaje.fromContext)
+          injectLogAnnotations *>
+            getCurrentContext.map(Baggaje.fromContext)
 
         override def get(name: String)(implicit trace: Trace): UIO[Option[String]] =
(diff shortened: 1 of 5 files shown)


opentelemetry/src/main/scala/zio/telemetry/opentelemetry/baggage/Baggage.scala
```diff
@@ -114,39 +114,44 @@ trait Baggage { self =>
 
 object Baggage {
 
-  def live: URLayer[ContextStorage, Baggage] =
+  def live(logAnnotated: Boolean = false): URLayer[ContextStorage, Baggage] =
```
grouzen, line 117: My biggest concern is how to encode several different capabilities. For Baggage, we likely wouldn't add more, but for Tracing, we're going to have the following:

- ZIO log annotations as Span attributes
- ZIO logging as Span log events

In this case, I chose to go the most straightforward way and added one more argument to the live layer constructor to dispatch on it. It leads to one additional boolean check in the runtime. Of course, for Baggage, we can have separate `Baggage` trait implementations because we have only one capability. In the case of Tracing, we might want to give a user an option to combine these capabilities the way he/she wants, and for this, the approach of having multiple implementations of the Tracing trait wouldn't work.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/baggage/Baggage.scala
```diff
@@ -155,24 +160,46 @@ object Baggage {
...
+        private def modifyContext(body: Context => Context)(implicit trace: Trace): UIO[Context] =
+          ctxStorage.updateAndGet(body)
+
+        private def injectLogAnnotations(implicit trace: Trace): UIO[Unit] =
+          ZIO
+            .when(logAnnotated) {
+              for {
+                annotations            <- ZIO.logAnnotations
+                annotationsWithMetadata = annotations.map { case (k, v) =>
+                                            (k, (v, BaggageEntryMetadata.create("zio log annotation")))
```
grouzen, line 187: Not sure if it's worth adding this metadata.

PR #681, opened 2023-05-06
Remove PropagatingSupervisor

dmytr:
Since [version 1.25.0](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v1.25.0) OpenTelemetry JVM agent supports ZIO. There is no need to use `PropagatingSupervisor` now.

This PR:
* Removes `PropagatingSupervisor`.
* Renames `ContextStorage.threadLocal` to `ContextStorage.openTelemetryContext` as it's a much more fitting name.
* Removes `Tracing.propagating` because now it's enough to use `Tracing.live` in combination with `ContextStorage.openTelemetryContext`.
* Updates docs section about usage with OpenTelemetry automatic instrumentation.

These are breaking changes, but since we are in RC phase, I hope this is acceptable.


grouzen (dismissed):
That's a significant improvement. Thank you!

grouzen (approved):

PR #695, opened 2023-05-24
OpenTelemetry: add support for recording exceptions in spans

### LINKED ISSUE #685: OpenTelemetry: add support for recording exceptions in spans
For more details, see the spec: https://opentelemetry.io/docs/instrumentation/java/manual/#record-exceptions-in-spans

IvanFinochenko:


IvanFinochenko:
ISSUE https://github.com/zio/zio-telemetry/issues/685

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
```diff
@@ -685,17 +691,22 @@ object Tracing {
...
+                .flatMap(errorMapper.body.lift)
                 .getOrElse(StatusCode.UNSET)
 
-            ZIO.succeed(span.setStatus(errorStatus, cause.prettyPrint))
+            for {
+              spanWithStatus <- ZIO.succeed(span.setStatus(errorStatus, cause.prettyPrint))
+              result         <- errorMapper.toThrowable.fold(ZIO.succeed(spanWithStatus))(toThrowable =>
+                                  ZIO.succeed(spanWithStatus.recordException(cause.squashWith(toThrowable)))
+                                )
+            } yield result
```
grouzen: Since `span.setStatus` is mutable, we might reuse `span` in the code below instead of introducing an additional `spanWithStatus` variable. Example:
```
            for {
              _      <- ZIO.succeed(span.setStatus(errorStatus, cause.prettyPrint))
              result <- errorMapper.toThrowable.fold(ZIO.succeed(span))(toThrowable =>
                          ZIO.succeed(span.recordException(cause.squashWith(toThrowable)))
                        )
            } yield result
```
or even 
```
            ZIO.succeed(span.setStatus(errorStatus, cause.prettyPrint)) *>
              errorMapper.toThrowable.fold(ZIO.succeed(span))(toThrowable =>
                ZIO.succeed(span.recordException(cause.squashWith(toThrowable)))
              )
```

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
```diff
@@ -685,17 +691,22 @@ object Tracing {
             getCurrentSpan.map(_.setAttribute(AttributeKey.doubleArrayKey(name), v))
           }
 
-          private def setErrorStatus[E](
+          override def setErrorStatus[E](
             span: Span,
             cause: Cause[E],
-            toErrorStatus: ErrorMapper[E]
+            errorMapper: ErrorMapper[E]
```
grouzen, line 715: I like this name. Could you please change every occurrence of `toErrorStatus` to `errorMapper` in the code base if you have time?

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
```diff
@@ -414,6 +414,12 @@ trait Tracing { self =>
     trace: Trace
   ): UIO[Span]
 
+  def setErrorStatus[E <: Throwable](
+    span: Span,
+    cause: Cause[E],
+    errorMapper: ErrorMapper[E]
+  )(implicit trace: Trace): UIO[Span]
+
```
grouzen: Why do you think we need to make it a part of the public API?

grouzen:
Hey @IvanFinochenko! Thank you for your contribution!

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
```diff
@@ -414,6 +414,12 @@ trait Tracing { self =>
     trace: Trace
   ): UIO[Span]
 
+  def setErrorStatus[E <: Throwable](
```
grouzen: The subtyping annotation `<: Throwable`  is redundant here.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/ErrorMapper.scala
```diff
@@ -2,11 +2,11 @@ package zio.telemetry.opentelemetry.tracing
 
 import io.opentelemetry.api.trace.StatusCode
 
-case class ErrorMapper[E](body: PartialFunction[E, StatusCode])
+case class ErrorMapper[E](body: PartialFunction[E, StatusCode], toThrowable: Option[E => Throwable])
```
grouzen: I think it would be better to make `toThrowable` equal to None by default.
```
case class ErrorMapper[E](body: PartialFunction[E, StatusCode], toThrowable: Option[E => Throwable] = None)
```

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/ErrorMapper.scala
```diff
@@ -2,11 +2,11 @@ package zio.telemetry.opentelemetry.tracing
...
 
 object ErrorMapper {
 
   def default[E]: ErrorMapper[E] =
-    ErrorMapper[E](Map.empty)
+    ErrorMapper[E](Map.empty, None)
```
grouzen: So here you don't need to provide `None` explicitily.

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/http/ProxyHttpApp.scala
```diff
@@ -1,21 +1,21 @@
...
-import zio.telemetry.opentelemetry.tracing.{ ErrorMapper, Tracing }
+import zio.telemetry.opentelemetry.tracing.{ErrorMapper, Tracing}
 
 case class ProxyHttpApp(client: Client, tracing: Tracing, baggage: Baggage) {
 
   import tracing.aspects._
 
   private val errorMapper: ErrorMapper[Throwable] =
-    ErrorMapper[Throwable] { case _ => StatusCode.UNSET }
+    ErrorMapper[Throwable]({ case _ => StatusCode.UNSET }, Some(identity))
```
grouzen: If we apply the change I suggested in the comment below https://github.com/zio/zio-telemetry/pull/695/files#r1203978344, the definition of this one would be the same as before:
```
  private val errorMapper: ErrorMapper[Throwable] =
    ErrorMapper[Throwable] { case _ => StatusCode.UNSET }
```
Which looks much cleaner from the end user's perspective.

grouzen:
LGTM! There is just a couple of comments to address. And also, it would be great to add proper tests for this new functionality.

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
IvanFinochenko: I made it public as in OpenTracing
https://github.com/zio/zio-telemetry/blob/series/2.x/opentracing/src/main/scala/zio/telemetry/opentracing/OpenTracing.scala#L17

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
IvanFinochenko: fixed

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
IvanFinochenko: changed

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
IvanFinochenko: fixed

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/ErrorMapper.scala
IvanFinochenko: fixed

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/ErrorMapper.scala
IvanFinochenko: fixed

opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/http/ProxyHttpApp.scala
IvanFinochenko: fixed

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
grouzen: I see. The thing is that opentracing module needs some love, and I'm planning to give it when we stabilize the API of opentelemetry, which is our playground for the moment. So please, do not refer to opentracing module as an example.

Considering the above I would like to ask you about making this method private. Thanks!

grouzen (changes requested):
Please add proper tests.

IvanFinochenko:
@grouzen, I added test.
But I have problems with some checks, can you help, please? What's wrong?
locally sbt compile is succeed.

IvanFinochenko:
@grouzen, I fixed it

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
IvanFinochenko: fixed

grouzen:
@IvanFinochenko Hey! Nice! I tried to figure out why it happened yesterday but with no success.

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/tracing/TracingTest.scala
```diff
@@ -123,7 +123,7 @@ object TracingTest extends ZIOSpecDefault {
               .flatMap(_.getAttributes.asMap().asScala.toList.map(x => x._1.getKey -> x._2.toString)),
             hasSubset(List("exception.message" -> "some_error", "exception.type" -> "java.lang.RuntimeException"))
           )
-          val errorMapper                       = ErrorMapper[Throwable](_ => StatusCode.ERROR, Some(identity))
+          val errorMapper                       = ErrorMapper[Throwable]({ case _ => StatusCode.ERROR }, Some(identity))
```
grouzen, line 126: A small thing - we don't  need `Some(identity)` anymore since we made this parameter None by default. 
Btw, nice catch! It is strange why the difference in the syntax causes the compiler to fail.

grouzen (approved):
Let's fix the errorMapper definition in tests and I'm happy to merge it! Good job!

opentelemetry/src/test/scala/zio/telemetry/opentelemetry/tracing/TracingTest.scala
IvanFinochenko: if I set None then I can't catch error and test will be failed

grouzen: Ah sorry, you're right!

PR #701, opened 2023-05-29
OpenTelemtry: set status description only when StatusCode is ERROR

### LINKED ISSUE #699: OpenTelemetry: change a default error status to ERROR instead of UNSET
Taken from https://github.com/zio/zio-telemetry/issues/444:

> Unless the operation contains an error, the StatusCode should remain "unset" and attempts to set value Unset will be ignored. A more sensible default would be to always set an Error StatusCode for any defect or failure and allow the caller to change if needed.

### LINKED ISSUE #700: OpenTelemtry: set status description only when StatusCode is ERROR
Taken from https://github.com/zio/zio-telemetry/issues/444:

> It is a bit confusing: According to the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status), Description is optional (for example: [Don't set the span status description](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status) if the reason can be inferred from http.status_code) and MUST only be used with the Error StatusCode value.

IvanFinochenko:


IvanFinochenko:
ISSUE https://github.com/zio/zio-telemetry/issues/699
ISSUE https://github.com/zio/zio-telemetry/issues/700

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
```diff
@@ -717,10 +717,13 @@ object Tracing {
...
                 .flatMap(errorMapper.body.lift)
-                .getOrElse(StatusCode.UNSET)
+                .getOrElse(StatusCode.ERROR)
 
             for {
-              _      <- ZIO.succeed(span.setStatus(errorStatus, cause.prettyPrint))
+              _      <- ZIO.ifZIO(ZIO.succeed(errorStatus == StatusCode.ERROR))(
+                          onTrue = ZIO.succeed(span.setStatus(errorStatus, cause.prettyPrint)),
+                          onFalse = ZIO.succeed(span.setStatus(errorStatus))
+                        )
```
grouzen: I suggest changing it to a simple if/else statement like this:
```
              _      <- if (errorStatus == StatusCode.ERROR)
                          ZIO.succeed(span.setStatus(errorStatus, cause.prettyPrint))
                        else
                          ZIO.succeed(span.setStatus(errorStatus))
```

grouzen:
LGTM! There is one minor comment to address only. Thanks for the contribution!

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala
IvanFinochenko: Ok, fixed

grouzen (approved):
Great job!

PR #709, opened 2023-06-04
OpenTelemetry: add Tracing.spanScoped method

### LINKED ISSUE #703: OpenTelemetry: add Tracing.spanScoped method
There is a feature request in the Discord channel (https://discord.com/channels/629491597070827530/639825316021272602/1113090265041862697) to add an ability to make a "scoped span".

So, we need to add a new method similar to the existing `Tracing.span`. It must not take a zio effect as an argument but instead return scoped ZIO value, like:
```
  def span(
    spanName: String,
    spanKind: SpanKind = SpanKind.INTERNAL,
    attributes: Attributes = Attributes.empty(),
    links: Seq[SpanContext] = Seq.empty
  )(implicit trace: Trace): ZIO[Scope, Nothing, Unit]
```
There is no need to add an aspect method since it doesn't wrap up zio effect.

Usage examples:

```
def makeStream(): ZStream[Any, Nothing, Int] =
  ZStream.scoped(tracing.spanScoped("stream span")) *> ZStream.repeat(1).take(10)
```

```
val workflow: ZIO[Scope, Nothing, String] = for {
  _ <- tracing.spanScoped("parent span name")
  _ <- tracing.addEvent("parent event")
  _ <- tracing.spanScoped("child span name") // It is worth checking if this "parent/child" semantics makes sense for spanScoped
  _ <- tracing.addEvent("child event")
  a <- ???
} yield a

val workflow1 = ZIO.scoped(ZIO.debug("workflow 1") *> workflow)
val workflow2 = ZIO.scoped(ZIO.debug("workflow 2") *> workflow)
```

oridag:
resolves #703

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/context/ContextStorage.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/context/ContextStorage.scala
@@ -15,6 +15,7 @@ trait ContextStorage {
 
   def locally[R, E, A](context: Context)(zio: ZIO[R, E, A])(implicit trace: Trace): ZIO[R, E, A]
 
+  def locallyScoped(context: Context)(implicit trace: Trace): ZIO[Scope, Nothing, Unit]
 }
 
 object ContextStorage {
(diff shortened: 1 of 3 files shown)


grouzen (approved):
Thanks for the contribution! It looks great!

PR #734, opened 2023-07-15
Improve usability

### LINKED ISSUE #733: OpenTelemetry: find a way to avoid specifying type parameters when calling Tracing aspects
It is quite annoying that we need to pay the price of typing useless type annotations each time we call `zio @@ span[Throwable, Unit]("...")` or similar aspect methods when initially the goal of introducing them was to have better API ergonomics. 
We ended up in a situation where some of the methods require typing 3 type parameters after https://github.com/zio/zio-telemetry/pull/702 gets merged.

regiskuckaertz:


grouzen (approved):
Amazing! Thanks for the contribution!

PR #786, opened 2023-11-24
Log a Throwable rather than a Cause in OpenTracing's logError

ghostdogpr:
Some tracer libraries such as dd-trace (Datadog) have special logic when the error type is a `Throwable`, but are not able to do anything with a `Cause`.

Before I would see this with no details about the error:
Screen Shot 2023-11-24 at 10 20 33 AM

After that change, I would get the error details with the full stack trace:
Screen Shot 2023-11-24 at 10 21 29 AM

The [spec](https://github.com/opentracing/specification/blob/master/semantic_conventions.md) says "For languages that support such a thing (e.g., Java, Python), the actual Throwable/Exception/Error object instance itself." so I think it makes sense not just for Datadog.

Relevant logic in dd-trace: https://github.com/DataDog/dd-trace-java/blob/a760c687369c971fa38d95c103883ed672394e6c/dd-trace-ot/src/main/java/datadog/opentracing/DefaultLogHandler.java#L45

### DIFF
--- a/opentracing/src/main/scala/zio/telemetry/opentracing/OpenTracing.scala
+++ b/opentracing/src/main/scala/zio/telemetry/opentracing/OpenTracing.scala
@@ -126,7 +126,9 @@ object OpenTracing {
         )(implicit trace: Trace): UIO[Unit] =
           for {
             _ <- ZIO.succeed(span.setTag("error", true)).when(tagError)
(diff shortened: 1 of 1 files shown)


grouzen:
Thanks for the contribution! It is definitely a reasonable change and I'm happy to merge it!

grouzen (approved):

ghostdogpr:
Thanks! Would it be possible to do a release in the near future? 🙏

grouzen:
Sure, I'll cut the next release tomorrow.

PR #795, opened 2024-01-12
[OpenTracing] Preserve Throwable failures when logging errors

ghostdogpr:
This is a follow-up to my previous PR: https://github.com/zio/zio-telemetry/pull/786

When the error is already a `Throwable`, it's a bit pity to wrap it in a `FiberFailure` because then all errors end up being `FiberFailure`. When we already have a `Throwable`, we don't need that wrapping.


grouzen (approved):
Thanks for this fix!

PR #824, opened 2024-04-23
Add mima checking

### LINKED ISSUE #750: Configure MIMA to fail the build on backward incompatible code changes.
Example: https://github.com/zio/zio-logging/blob/master/project/MimaSettings.scala

IvanFinochenko:
Issue https://github.com/zio/zio-telemetry/issues/750

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,15 +100,15 @@ jobs:
       - name: Install libuv
         run: sudo apt-get update && sudo apt-get install -y libuv1-dev
       - name: Setup Scala
-        uses: actions/setup-java@v3.10.0
+        uses: actions/setup-java@v4.2.1
         with:
-          distribution: temurin
-          java-version: "8"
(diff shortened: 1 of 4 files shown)


grouzen:
@IvanFinochenko Hello! Thanks for the contribution! 

A few questions/suggestions. 
I believe this PR is missing the CI part to run the check, which is a `mimaCheck` alias you've added.
This can't be done via amending `ci.yml` directly since we're using [sbt-zio plugin](https://zio.dev/zio-sbt/) that generates it for us. It means we must use one of the available sbt-zio directives to include the mimaCheck command in the CI pipeline.
After doing this, we just need to set `mimaFailOnProblem` to false until we release the `3.0.0` version.
That's just how I see it. I hope it makes sense.

IvanFinochenko:
@grouzen Hello! Thanks for suggestions.
I have fixed it.

grouzen (approved):

Issue #832, opened 2024-05-08
Opentelemetry-example doesn't collect data to Jaeger and Seq


IvanFinochenko:
I run scripts from opentelemetry-example.md, open Jaeger UI and Seq UI and don't see any data.

grouzen:
Did you change the providers? 
There are different tracing and logging providers available:
- https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/otel/TracerProvider.scala
- https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/otel/LoggerProvider.scala

By default, we use stdout for the examples.
There is a line in the doc that explains this: https://github.com/zio/zio-telemetry/blob/series/2.x/docs/opentelemetry-example.md?plain=1#L35
Perhaps the documentation regarding this part is not written in the best way.

IvanFinochenko:
@grouzen Thanks for answer. I have opened PR with some fix

PR #833, opened 2024-05-08
Fix opentelemetry example

IvanFinochenko:
Issue https://github.com/zio/zio-telemetry/issues/832

### DIFF
--- a/docs/opentelemetry-example.md
+++ b/docs/opentelemetry-example.md
@@ -62,4 +62,16 @@ docker run \
   datalust/seq
 ```
 
-Run the application and fire a curl request as shown above. Head over to [Jaeger UI](http://localhost:16686/) and [Seq UI](http://localhost:80/) to see the result.
+Run proxy:
+```bash
+sbt "opentelemetryExample/runMain zio.telemetry.opentelemetry.example.ProxyApp jaeger-seq"
+```
+Run backend:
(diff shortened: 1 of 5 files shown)


grouzen:
Hey! 
This is a good catch.
But I doubt we need to merge this. I think we shouldn't overcomplicate the examples. The better option would be to add explicit instructions on changing the OtelSdk to use Jaeger and Seq providers instead of default ones. This approach will encourage people to play with examples' code to gain a better understanding. 
For this, we could just point to the specific lines in https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/otel/TracerProvider.scala#L37 and https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/otel/LoggerProvider.scala#L36, saying that a user needs to replace the stdout providers in https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry-example/src/main/scala/zio/telemetry/opentelemetry/example/otel/OtelSdk.scala#L13 with the above ones.
What do you think?

IvanFinochenko:
Yes, I agree. I changed it new PR https://github.com/zio/zio-telemetry/pull/835

PR #864, opened 2024-07-30
Pass ZIO Metrics histogram boundaries to OTEL

andrzejressel:
Currently boundaries are ignored and default OTEL ones are used.


grouzen (approved):
Thanks for your contribution! Great stuff!

PR #871, opened 2024-08-09
Remove Double.MaxValue histogram upper bound added by ZIO

### LINKED ISSUE #870: OTEL histogram contains two infs
When using boundaries fromChunk, Double.MaxValue is added to the end. But OTEL already adds inf. This causes two inf - one is native "inf" and second is just boundary with Double.MaxValue.

Test works because I've used default constructor. It should be changed to fromChunk factory instead.

While histogram can be workarounded by using default Histogram constructor, timer cannot be fixed.

andrzejressel:
Fixes #870

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/metrics/internal/Instrument.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/metrics/internal/Instrument.scala
@@ -102,7 +102,7 @@ object Instrument {
 
             unit.foreach(builder.setUnit)
             description.foreach(builder.setDescription)
-            boundaries.foreach(seq =>
+            boundaries.foreach { seq =>
               builder.setExplicitBucketBoundariesAdvice(
                 seq
                   // ZIO uses explicit Double.MaxValue boundary as upper bound. OTEL uses internal "inf" value instead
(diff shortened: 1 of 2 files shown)


opentelemetry/src/main/scala/zio/telemetry/opentelemetry/metrics/internal/Instrument.scala
```diff
@@ -102,7 +102,15 @@ object Instrument {
...
-            boundaries.foreach(seq => builder.setExplicitBucketBoundariesAdvice(seq.map(Double.box).asJava))
+            boundaries.foreach(seq =>
+              builder.setExplicitBucketBoundariesAdvice(
+                seq
+                  // ZIO uses explicit Double.MaxValue boundary as upper bound. OTEL uses internal "inf" value instead
+                  .filter(_ != Double.MaxValue)
+                  .map(Double.box)
+                  .asJava
+              )
+            )
```
grouzen: Nitpick: for the sake of syntactic consistency

```suggestion
            boundaries.foreach { seq =>
              builder.setExplicitBucketBoundariesAdvice(
                seq
                  // ZIO uses explicit Double.MaxValue boundary as upper bound. OTEL uses internal "inf" value instead
                  .filter(_ != Double.MaxValue)
                  .map(Double.box)
                  .asJava
              )
            }
```

grouzen (dismissed):
Great! Thanks for catching this annoying tidbit!

grouzen (approved):

Issue #911, opened 2024-11-18
Impossible to use `Attributes` in span because of Scala error: `Ambiguous overload. The overloaded alternatives of method apply in object Attributes with types`


jivanic-demystdata:
```scala
ZIO.service[Tracing].flatMap { tracing =>
  import tracing.aspects.*

  ZIO.attempt { ... ) @@ root(
    "my-trace", 
    attributes = Attributes(
      Attribute(AttributeKey.stringKey("id"), "1"),
      Attribute(AttributeKey.stringKey("example"), "value")
    )
}
```
fails to compile with:
```scala
Ambiguous overload. The overloaded alternatives of method apply in object Attributes with types
 [T, U]
  (attribute1: zio.telemetry.opentelemetry.common.Attribute[T], attribute2:
    zio.telemetry.opentelemetry.common.Attribute[U]):
    io.opentelemetry.api.common.Attributes
 [T]
  (attributes: zio.telemetry.opentelemetry.common.Attribute[T]*):
    io.opentelemetry.api.common.Attributes
both match arguments (zio.telemetry.opentelemetry.common.Attribute[String], zio.telemetry.opentelemetry.common.Attribute[String]) [89:28]
```

grouzen:
Thanks for reporting!

grouzen:
@jivanic-demystdata What scala version are you using? 

Also, could you try using the `Attribute.string` helper method (https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/common/Attribute.scala#L20) instead of wrapping a Java method `AttributerKey.stringKey`?

jivanic-demystdata:
@grouzen I'm using Scala 3.5.2

Same error when using `Attribute.string`:
```scala
Ambiguous overload. The overloaded alternatives of method apply in object Attributes with types
 [T, U]
  (attribute1: zio.telemetry.opentelemetry.common.Attribute[T], attribute2:
    zio.telemetry.opentelemetry.common.Attribute[U]):
    io.opentelemetry.api.common.Attributes
 [T]
  (attributes: zio.telemetry.opentelemetry.common.Attribute[T]*):
    io.opentelemetry.api.common.Attributes
both match arguments (zio.telemetry.opentelemetry.common.Attribute[String], zio.telemetry.opentelemetry.common.Attribute[String]) [88:28]
```

grouzen:
[...]

PR #913, opened 2024-11-25
Fix Attributes.apply methods resolution

grouzen:
Addresses #911 

This fix causes binary incompatibility. I think it is acceptable to release it as a minor version though, because it is a very minor change and shouldn't be a big deal for most users to adjust their code.


grouzen:
@jivanic-demystdata Could you please review it?

jivanic-demystdata (approved):

PR #1008, opened 2025-07-01
Add OpenTelemetry.zioLogger layer

### LINKED ISSUE #1007: OpenTelemetry: Provide a way of accessing the OTEL LogRecords ZLogger directly
I would like to use [Logging's](https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/logging/Logging.scala) ZLogger instance for small parts of my application. The reason for doing this is that the logs sent to the OTEL logger will be sent as New Relic Custom Events. By default, I would like my application to use the default configured ZLogger which goes to standard out.

I believe this is possible by making this public
```def zioLogger(instrumentationScopeName: String)...` 

And leveraging `ZIO.withLogger(Scoped)` in key parts of the application where you want those logs to be become OTEL LogRecords -> New Relic Custom events.

Is that a reasonable ask?

### PR DESCRIPTION
**Pull request by grouzen**
Addresses #1007

### DIFF
--- a/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/OpenTelemetry.scala
+++ b/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/OpenTelemetry.scala
@@ -237,6 +237,22 @@ object OpenTelemetry {
     }
 
   /**
+   * Returns an instance of `ZIO.ZLogger` that is configured to propagate log records as OTEL Log signals.
+   *
+   * It may be useful when you need to execute an effect with the specified logger. See `ZIO.withLogger`.
+   */
+  def zioLogger(
+    instrumentationScopeName: String
+  ): URLayer[OpenTelemetry, ZLogger[String, Unit]] =
+    ZLayer.scoped {
+      for {
+        openTelemetry <- ZIO.service[OpenTelemetry]
+        loggerProvider = openTelemetry.unsafe.asJava.getLogsBridge
+        logger         = Logger.zioLogger(instrumentationScopeName)(openTelemetry.ctxStorage, loggerProvider)
+      } yield logger
+    }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: opentelemetry/src/main/scala/zio/telemetry/opentelemetry/OpenTelemetry.scala [hunk 1]
```diff
@@ -236,6 +236,22 @@ object OpenTelemetry {
...
+    instrumentationScopeName: String
+  ): URLayer[OpenTelemetry, ZLogger[String, Unit]] =
+    ZLayer.scoped {
+      for {
+        openTelemetry <- ZIO.service[OpenTelemetry]
+        loggerProvider = openTelemetry.unsafe.asJava.getLogsBridge
+        logger         = Logger.zioLogger(instrumentationScopeName)(openTelemetry.ctxStorage, loggerProvider)
+      } yield logger
+    }
+
```
calvinlfer, line 254: Awesome! this is exactly what i need

calvinlfer (approved):

opentelemetry/src/main/scala/zio/telemetry/opentelemetry/logs/Logger.scala
```diff
@@ -22,7 +22,7 @@ private[opentelemetry] object Logger {
       _      <- ZIO.withLoggerScoped(logger)
     } yield ()
 
-  private def zioLogger(instrumentationScopeName: String)(
+  private[opentelemetry] def zioLogger(instrumentationScopeName: String)(
```
calvinlfer, line 25: thanks for making this package private!

calvinlfer (approved):

PR #1107, opened 2025-12-16
add support for `Span.setAllAttributes`

dontgitit:
https://discord.com/channels/629491597070827530/639825316021272602/1448193364234342542

Adds support for `Span.setAllAttributes`

### DIFF
--- a/opentelemetry-core/src/main/scala/zio/telemetry/opentelemetry/core/trace/Span.scala
+++ b/opentelemetry-core/src/main/scala/zio/telemetry/opentelemetry/core/trace/Span.scala
@@ -1,10 +1,10 @@
 package zio.telemetry.opentelemetry.core.trace
 
-import io.opentelemetry.api.common.{AttributeKey, Attributes}
+import io.opentelemetry.api.common.{AttributeKey, Attributes => JAttributes}
 import io.opentelemetry.api.trace.{Span => JSpan, SpanContext, StatusCode}
 import io.opentelemetry.context.Context
 import zio._
-import zio.telemetry.opentelemetry.core.common.Attribute
+import zio.telemetry.opentelemetry.core.common.{Attribute, Attributes}
 
 import java.util.concurrent.TimeUnit
 import scala.jdk.CollectionConverters._
@@ -33,7 +33,7 @@ trait Span { self =>
    */
   def addEventWithAttributes(
     name: String,
-    attributes: Attributes
+    attributes: JAttributes
   )(implicit trace: Trace): UIO[Unit]
 
   /**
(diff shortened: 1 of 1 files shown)


opentelemetry-core/src/main/scala/zio/telemetry/opentelemetry/core/trace/Span.scala
```diff
@@ -157,6 +157,11 @@ trait Span { self =>
     trace: Trace
   ): UIO[Unit]
 
+  def setAllAttributes(attributes: JAttributes)(implicit trace: Trace): UIO[Unit]
+
+  def setAllAttributes(attributes: List[Attribute[_]])(implicit trace: Trace): UIO[Unit] =
+    setAllAttributes(Attributes.fromList(attributes))
```
dontgitit: not sure if this is worth having, but thought i'd add it while i'm here...

grouzen: Not sure too. But it makes me think what if we replace all the flavours of `setAttribute` with just one `def setAttribute[T](attribute: Attribute[T])(implicit trace: Trace): UIO[Unit]` that receives the scala wrapper `common.Attribute[T]` (https://github.com/zio/zio-telemetry/blob/v4.0.0-rc/opentelemetry-core/src/main/scala/zio/telemetry/opentelemetry/core/common/Attribute.scala)

dontgitit: i don't have a strong opinion, though i don't really understand the motivation behind that replacement; i think the current version is nicer since you don't have to explicitly create an `Attribute`. To me, `setAttribute(key, value)` reads a bit nicer than `setAttribute(Attribute(key, value))`

grouzen: Yeah, I think I wouldn't add the method in question, and better force a user to create `JAttributes` via `Attributes.fromList` or `Attributes.apply` methods. 
This will reduce unnecessary redundancy in the current API. Additionally, we can remove all specific `setAttribute` methods except the generic one.

grouzen: > i don't have a strong opinion, though i don't really understand the motivation behind that replacement; i think the current version is nicer since you don't have to explicitly create an `Attribute`. To me, `setAttribute(key, value)` reads a bit nicer than `setAttribute(Attribute(key, value))`

yeah, this is true. But in case of `setAllAttributes` I would stick with `setAllAttributes(Attributes.fromList(Attribute(key, value), ...))` rather than with `setAllAttributes(List(Attribute(key, value), ...))`

dontgitit: removed it!

grouzen (approved):
LGTM! Thanks for another great contribution!

Issue #1022, opened 2025-07-24
Opentelemetry: seamless integration with ZIO.logSpan and ZIO.logAnnotated for creating OTEL spans with attributes


thiloplanz:
ZIO has a built-in `logSpan` method that can be used to mark spans. Those will end up as (timed) annotations on ZIO log messages within that span.

Are those in any way picked up by zio-telemetry? Or is there a manual way to collect that? So that for basic use-cases (just naming spans without further attributes) telemetry spans could be created without having to inject an extra service into application code. Just like we can do for metrics and logs (for which zio-opentelemetry provides bridging) ?

grouzen:
Hi! I had the same idea in the past, but I decided it was not worth it. 
Firstly, the name `logSpan` implies it is intended for logs, and additionally, it affects the rendering of ZIO logs as well.
Secondly, I don't think there is a simple way of creating OTEL spans from ZIO log spans, considering that the only way to collect the log spans is to call `ZIO.logSpans`, which returns a list of spans for the **current scope**. All this gives me the impression that we can't easily restore the hierarchy of nested spans out of it.
Additionally, we need to ensure it interoperates with `tracer.span` calls somehow. 

I hope it all makes sense!

I'm okay with adding this kind of bridge as you think it's doable and won't significantly impact performance.

thiloplanz:
> adding this kind of bridge as you think it's doable and won't significantly impact performance.

I agree with your assessment that there does not seem to be a good hook to intercept `logSpan` right now.

Logging and metrics have been integrated into ZIO core now, with the approach of providing a platform interface for provider libraries to plug into. I was hoping that since `logSpan` is a method present now that it would somehow also offer that (I cannot see why marking spans should be limited to log output, a span is a span). 

Do you think it would be a good thing to try to lobby for some hooks into ZIO to expose those logSpans? I guess that would mean registering a Tracer in the Runtime, just like we do with Logging now. Does it make sense? It would probably be limited by only being able to pass span names, not all the other extra attributes you could have.

A related question and possibly alternative approach: I yet have to look at auto-instrumentation via the Java agent. That one does have ZIO support. Do you know what it does (what it contributes to telemetry)? Does it create spans? If so, how?

grouzen:
[...]

PR #1138, opened 2026-02-17
feat: LogSpanner — pluggable span dispatch bridging ZIO.logSpan with OTEL (#1022)

li-nkSN:
Implements the LogSpanner design proposed in #1022: a FiberRef-based
dispatch mechanism that lets library code mark spans via
`effect @@ LogSpanner.span("name")` without depending on Tracer.

Three backends:
- Default: delegates to ZIO.logSpan (zero OTEL overhead)
- OTEL: creates real OpenTelemetry spans via Tracer.span
- Hybrid: creates both OTEL and ZIO logSpans

Includes 15 test cases covering default/OTEL/hybrid backends,
scoped installation, and fiber correctness (fork, timeout,
child fiber inheritance).

Design based on the prototype by @thiloplanz in #1022.
Thanks to @grouzen for design guidance and confirming the
ZIO.logAnnotate attribute propagation approach.

### DIFF
--- a/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/LogSpanner.scala
+++ b/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/LogSpanner.scala
@@ -1,0 +1,118 @@
+package zio.telemetry.opentelemetry.core.trace
+
+import zio._
+
+/**
+ * A pluggable span dispatch mechanism that bridges ZIO's `logSpan` with OpenTelemetry spans.
+ *
+ * By default, `LogSpanner.span("name")` delegates to `ZIO.logSpan`, adding zero OTEL overhead. When an OTEL-backed
+ * `LogSpanner` is installed (via `LogSpanner.installOtel` or `LogSpanner.installHybrid`), the same call creates real
+ * OTEL spans instead.
+ *
+ * This allows library code to use `LogSpanner.span` without depending on `Tracer`, while application code installs the
+ * OTEL backend at the edge.
+ *
+ * Implements the design proposed in zio-telemetry #1022.
+ */
+trait LogSpanner {
+
+  /**
+   * Wraps an effect with a named span.
+   *
+   * The semantics depend on the installed backend:
+   *   - Default: delegates to `ZIO.logSpan` (ZIO log-based spans only)
+   *   - OTEL: creates a real OpenTelemetry span via `Tracer.span`
+   *   - Hybrid: creates both an OTEL span and a ZIO logSpan
+   *
+   * @param name
+   *   the span name
+   * @param zio
+   *   the effect to wrap
+   */
+  def logSpan[R, E, A](name: String)(zio: => ZIO[R, E, A])(implicit trace: Trace): ZIO[R, E, A]
+}
+
+object LogSpanner {
+
+  /**
+   * Default implementation that delegates to `ZIO.logSpan`. Zero OTEL overhead — no spans are exported.
+   */
+  private[opentelemetry] val default: LogSpanner = new LogSpanner {
+
+    override def logSpan[R, E, A](name: String)(zio: => ZIO[R, E, A])(implicit trace: Trace): ZIO[R, E, A] =
+      ZIO.logSpan(name)(zio)
+  }
+
+  /**
+   * Global FiberRef storing the current LogSpanner.
+   *
+   * Uses the same `Unsafe.unsafe { FiberRef.unsafe.make }` pattern as `FiberRef.currentLogSpan` in ZIO core. This
+   * ensures the FiberRef is available at module initialization time without requiring a ZIO runtime.
+   *
+   * IMPORTANT: `default` must be defined before this val to avoid null initialization.
+   */
+  private[opentelemetry] val currentLogSpanner: FiberRef[LogSpanner] =
+    Unsafe.unsafe { implicit unsafe =>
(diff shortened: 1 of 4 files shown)


grouzen:
Hello! Great stuff! 

I would like to suggest streamlining the resulting design to align with the library's established architecture.
Please have a look at the draft PR I created in here: https://github.com/zio/zio-telemetry/pull/1151/changes.

It basically just moves the constructors and layers around to make it more coherent with the current design. 
Also, it adds ZIO's `implicit trace: Trace` parameter to the methods and layers.

I haven't touched the tests in my PR, so they need to be adjusted accordingly if you are happy with the proposed changes. 

Feel free to copy or re-create the suggested changes in your PR so I can remove my draft PR later.

docs/logspanner-design-and-usage.md
```diff
@@ -0,0 +1,437 @@
+# LogSpanner: Bridging ZIO.logSpan with OpenTelemetry Spans
```
grouzen: Sorry, this guy needs to be removed. 
I'm planning to update the documentation later when we understand that version 4 is ready to be released. So please do not bother updating the docs. Just please make sure that the scaladoc is fine.

modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/LogSpanner.scala
```diff
@@ -0,0 +1,118 @@
...
+   * @param spanName
+   *   the span name
+   * @return
+   *   a ZIOAspect that applies the span
+   */
+  def span(spanName: String): ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] =
+    new ZIOAspect[Nothing, Any, Nothing, Any, Nothing, Any] {
+      override def apply[R, E, A](zio: ZIO[R, E, A])(implicit trace: Trace): ZIO[R, E, A] =
+        currentLogSpanner.getWith(_.logSpan(spanName)(zio))
+    }
```
grouzen, line 117: I probably missed mentioning this one in my PR, but it can be removed since we basically moved it to `OpenTelemetry.logSpan`. See the next comment below

modules/opentelemetry/main/src/test/scala/zio/telemetry/opentelemetry/core/trace/LogSpannerTest.scala
```diff
@@ -0,0 +1,278 @@
...
+  // ---------------------------------------------------------------------------
+  // Default backend tests
+  // ---------------------------------------------------------------------------
+
+  private val defaultBackendSuite =
+    suite("default backend")(
+      test("delegates to ZIO.logSpan — no OTEL span created") {
+        for {
+          tracerTestkit <- ZIO.service[TracerTestkit]
+          _             <- ZIO.unit @@ LogSpanner.span("mySpan")
```
grouzen, line 48: Sorry, I had to be explicit about this. The idea is to replace this call with `OpenTelemetry.logSpan`.

This way, we kinda mimic the `ZIO.logSpan`, using the same naming convention.

li-nkSN:
@grouzen    Thanks for the restructuring feedback — we've incorporated all of it (curried signatures, implicit trace, installOtel/installHybrid on LogSpanner, OpenTelemetry.logSpan + aspects.logSpan on the trait, layer factories on the OpenTelemetry object).

  Regarding removing LogSpanner.span — we'd like to keep it alongside OpenTelemetry.aspects.logSpan. The reason is the viral dependency problem that #1022 was designed to solve.

  Consider a database module that instruments repository operations:

  // database module — depends only on zio-opentelemetry-core
  import zio.telemetry.opentelemetry.core.trace.LogSpanner

  def findById(id: Long): Task[Option[Entity]] =
    repository.findById(id) @@ LogSpanner.span("repo:findById")

  No OpenTelemetry instance needed. No environment requirement. The module depends only on zio-opentelemetry-core.

  If LogSpanner.span is removed and the call site must use openTelemetry.aspects.logSpan:

  // database module — now needs OpenTelemetry instance
  def findById(id: Long): Task[Option[Entity]] =
    repository.findById(id) @@ openTelemetry.aspects.logSpan("repo:findById")

  Where does openTelemetry come from? Either a constructor parameter or ZIO.serviceWithZIO[OpenTelemetry]. Both propagate — every module that calls findById must provide OpenTelemetry, and every module that calls those modules must provide it too. This is the viral Tracer dependency that
  #1022's FiberRef design specifically eliminates.

  LogSpanner.span and OpenTelemetry.aspects.logSpan are functionally identical — same FiberRef[LogSpanner], same dispatch. They can coexist: OpenTelemetry.aspects.logSpan for application code that already has an instance, LogSpanner.span for library/infrastructure code that shouldn't need
  one.

grouzen:
Hi! Sorry for being silent for so long. 

I've thought about it a lot, and I feel like this kind of stuff shouldn't be added to the library in this form. 

I think the proper approach would be to add a similar mechanism to the one used for switching between logging backends in ZIO: https://zio.dev/guides/tutorials/create-custom-logger-for-a-zio-application. I believe, this is the conclusion we came to with @thiloplanz in the original issue. The main idea is we want to be able to swap the backends of logSpan and logAnnotate using the non-existent ZIO API.

The main argument against the suggested API is that it feels like we are adding another clunky version of two already existing APIs:
- zio-telemetry: tracer.span and tracer.addAttribute
- ZIO: ZIO.logAnnotate and ZIO.logSpan

The better strategy would be to keep this additional layer/API in a separate library.

Issue #1069, opened 2025-10-21
[OpenTelemetry/Tracing] Interrupted zio effects are not properly traced.


fredDJSonos:
When using `zio.telemetry.opentelemetry.tracing` (`3.1.10`), interrupted zio effect are not recorded correctly.

Depending on the way we record a span, the interruption of the fiber can be totally ignored or just not ideally reported:
* when we use [spanScoped](https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala#L625), at least an error situation is reported because the implementation is using `ZIO.acquireReleaseExit` that calls the releaser on interruption.  ~~But it would be even better to have a hint of what’s going with  `otel.status_description == "interrupted"`~~.  UPDATE: there is something in `otel.status_description` in the case of an interruption (`cause.prettyPrint`).  It is good enough.
* when we use [span](https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala#L591) or anything else that relies on [finalizeSpanUsingEffect](https://github.com/zio/zio-telemetry/blob/series/2.x/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/tracing/Tracing.scala#L802), nothing is reported in the span.  `finalizeSpanUsingEffect` is using `tapErrorCause`, but it looks like `tapErrorCause` does not catch an interruption.

By the way, the same problem is presumably in [v4](https://github.com/zio/zio-telemetry/blob/v4.0.0-rc/opentelemetry/src/main/scala/zio/telemetry/opentelemetry/trace/Tracer.scala#L382) since `.exit` does not intercept an interruption either.

grouzen:
Hello! Thanks for reporting this. We've never thought about handling the interruption properly. Moreover, the OTEL standard doesn't address it for the time being.

I think the common workaround is to set the status to `ERROR` and the description to `interrupted` as you suggested.


[...]

PR #1139, opened 2026-02-17
fix: ensure interrupted spans record correct status via uninterruptibleMask (partial #1069)

li-nkSN:
This addresses Sub-problem A of #1069: interrupted spans are recorded
with UNSET status instead of ERROR because the interruption signal
races with statusMapper.handle in Tracer.contextScope.

Fix: wrap contextScope's exit/statusMapper sequence in
ZIO.uninterruptibleMask so that statusMapper.handle runs before the
interruption propagates. This is a ~4-line change.

This does NOT address Sub-problem B of #1069: ThreadLocal context
loss when fibers hop threads (affects OpenTelemetry.global / Java
agent auto-instrumentation path only). Sub-problem B requires a
FiberRef.asThreadLocal bridge — a substantially larger effort that
is orthogonal to this fix.

Adds InterruptionTracerTest verifying:
- FiberRef ContextStorage: timeout ERROR status, parent context
  preservation, race, fork, forkDaemon (5 tests, all passing)
- ThreadLocal ContextStorage: documents known #1069 behavior with
  @@ TestAspect.flaky(3) as regression tests for a future fix
  (2 tests)

Thanks to @dontgitit for reporting the timeout context loss in
#1069, and to @grouzen for the discussion on .onExit vs
uninterruptibleMask approaches.

### DIFF
--- a/modules/opentelemetry/main/src/test/scala/zio/telemetry/opentelemetry/core/trace/InterruptionTracerTest.scala
+++ b/modules/opentelemetry/main/src/test/scala/zio/telemetry/opentelemetry/core/trace/InterruptionTracerTest.scala
@@ -1,0 +1,147 @@
+package zio.telemetry.opentelemetry.core.trace
+
+import io.opentelemetry.api.trace.StatusCode
+import zio._
+import zio.telemetry.opentelemetry.testkit.OpenTelemetryTestkit
+import zio.telemetry.opentelemetry.testkit.trace.{SpanData, TracerTestkit}
+import zio.test.Assertion._
+import zio.test._
+
+/**
+ * Tests verifying that OTEL spans are correctly recorded when effects are interrupted.
+ *
+ * These tests document the behavior described in zio-telemetry #1069. The FiberRef-based ContextStorage path (used by
+ * `OpenTelemetry.custom`) is expected to handle interruption correctly because `Tracer.span` uses
+ * `ZIO.acquireReleaseWith`, which per Zionomicon Ch.14 guarantees that the release (endSpan) runs even on interruption.
+ *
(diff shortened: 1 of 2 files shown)


modules/opentelemetry/main/src/test/scala/zio/telemetry/opentelemetry/core/trace/InterruptionTracerTest.scala
```diff
@@ -0,0 +1,147 @@
...
+   * FiberRef ContextStorage tests — these should all PASS on RC10, confirming that #1069 is a ThreadLocal-only issue
+   * for the `acquireReleaseWith` span lifecycle path.
+   */
+  private val fiberRefSuite =
+    suite("FiberRef ContextStorage")(
+      test("timeout: span is recorded with ERROR status when effect times out") {
+        for {
+          tracerTestkit <- ZIO.service[TracerTestkit]
+          tracer        <- tracerTestkit.getTracer(instrumentationScopeName)
+          _             <- tracer.span("slow-op")(_ => ZIO.sleep(1.hour)).timeout(10.millis)
```
grouzen: suggestion: you can avoid skipping the `span` parameter in the closure by using `tracer.aspects.span` instead of `tracer.span`. 
Example:
```
(ZIO.sleep(1.hour) @@ tracer.aspects.span("slow-op")).timeout(10.millis)
```

I can see that the same pattern applies to the rest of the `tracer.span` calls in this test suite.

grouzen (dismissed):
LGTM! 

There is a minor comment to simplify the test code. Apart from this, great work!

grouzen (approved):
LGTM!

PR #1215, opened 2026-08-06
add `Cause` variants to `StatusMapper` that allow acting on defects

dontgitit:
when using the `StatusMapper` API, there was previously no way to record exceptions for defects (or extract stack traces from failures, unless the failure was itself a throwable); it had to be done via an explicit `recordException` using something like this: `.tapErrorCause(e => span.recordException(...))`

additionally, it was impossible to set a description based on a defect, because even if you explicitly call `span.setStatus(status, description)`, the `StatusMapper` would overwrite the description (with `""`) when the span ends.

for more discussion, see [this thread on Discord](https://discord.com/channels/629491597070827530/639825316021272602/1526684761098158131)

### DIFF
--- a/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
+++ b/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
@@ -43,7 +43,7 @@ object StatusMapper {
   }
 
   /**
-   * Default case allows overriding the default behavior from the official specfication:
+   * Default case allows overriding the default behavior from the official specification:
    * [[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status Set status]]
    *
    * @param success
@@ -51,7 +51,7 @@ object StatusMapper {
    */
   sealed class Default[-E, -A](
     success: PartialFunction[A, Result.Success] = PartialFunction.empty,
-    failure: PartialFunction[E, Result.Failure] = PartialFunction.empty
+    failure: PartialFunction[Cause[E], Result.Failure] = PartialFunction.empty
   ) extends StatusMapper[E, A] {
 
     override def handleSuccess(span: Span, a: A)(implicit trace: Trace): UIO[Unit] =
@@ -68,8 +68,8 @@ object StatusMapper {
 
     override def handleFailure(span: Span, cause: Cause[E])(implicit trace: Trace): UIO[Unit] = {
       val result =
-        cause.failureOption
-          .flatMap(failure.lift)
+        failure
+          .lift(cause)
           .getOrElse(StatusMapper.Result.Failure(StatusCode.ERROR))
 
       for {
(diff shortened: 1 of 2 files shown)


modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
```diff
@@ -43,15 +43,15 @@ object StatusMapper {
...
+   * Default case allows overriding the default behavior from the official specification:
    * [[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status Set status]]
    *
    * @param success
    * @param failure
    */
   sealed class Default[-E, -A](
     success: PartialFunction[A, Result.Success] = PartialFunction.empty,
-    failure: PartialFunction[E, Result.Failure] = PartialFunction.empty
+    failure: PartialFunction[Cause[E], Result.Failure] = PartialFunction.empty
```
dontgitit, line 54: we now accept a partial over `Cause[E]` rather than just `E`

modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
```diff
@@ -68,8 +68,8 @@ object StatusMapper {
 
     override def handleFailure(span: Span, cause: Cause[E])(implicit trace: Trace): UIO[Unit] = {
       val result =
-        cause.failureOption
-          .flatMap(failure.lift)
+        failure
+          .lift(cause)
```
dontgitit, line 72: no longer call `cause.failureOption`; the `failure` partial is responsible for doing that if it wants to handle only failures vs defects

modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
```diff
@@ -180,11 +184,29 @@ object StatusMapper {
    */
   def failure[E](toStatusCode: E => StatusCode)(toDescription: E => Option[String])(
     toException: E => Option[Throwable]
+  ): Failure[E] =
+    failureFromZIOFailure(e => Result.Failure(toStatusCode(e), toDescription(e), toException(e)))
```
dontgitit, line 188: the existing `failure` method is now where `cause.failureOption` is called (via `failureFromZIOFailure` above)

modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
```diff
@@ -224,7 +280,7 @@ object StatusMapper {
    * @return
    */
   def failureThrowable(toStatusCode: Throwable => StatusCode): Failure[Throwable] =
```
dontgitit, line 282: i didn't add any `Cause` equivalent for `failureThrowable`

grouzen:
This is brilliant! Thank you for the contribution!

grouzen (approved):

PR #1223, opened 2026-08-11
feat: configure default `StatusMapper` when creating `Tracer`

dontgitit:
The default behavior sets `ERROR` status for spans, but omits exception traces. This makes it difficult to debug issues. 
This PR aims to make it easy to configure the default behavior.

see more discussion here: https://discord.com/channels/629491597070827530/639825316021272602/1526684761098158131

follow-up to https://github.com/zio/zio-telemetry/pull/1215

### DIFF
--- a/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
+++ b/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/StatusMapper.scala
@@ -141,11 +141,17 @@ object StatusMapper {
     new Empty
 
   /**
-   * It follows the offical specification:
+   * It follows the official specification:
    * [[https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status Set status]]
    */
   val default: Default[Any, Any] =
     new Default()
+
+  /**
+   * Same as [[default]] above, but also sets exception stack traces
+   */
+  val defaultRecordingExceptions: Failure[Any] =
+    StatusMapper.failureCauseNoDescription[Any](_ => StatusCode.ERROR)(c => Some(FiberFailure(c)))
 
   /**
    * Overrides both success and failure cases of the default status mapper.
--- a/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/Tracer.scala
+++ b/modules/opentelemetry/core/src/main/scala/zio/telemetry/opentelemetry/core/trace/Tracer.scala
@@ -11,6 +11,8 @@ import scala.concurrent.ExecutionContext
 
 trait Tracer { self =>
 
+  protected val defaultStatusMapper: StatusMapper[Any, Any]
+
   /**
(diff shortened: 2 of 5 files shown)


grouzen (approved):
I think it is a great idea! It feels much better than a bunch of feature flags because it uses the same status mapper DSL and has the same flexibility.

Sorry, it took me some time to review because I was trying to figure out how to avoid recoding the same exception multiple times when it happens in one of the deeply nested spans. But I found that this is common across OTEL implementations in different programming languages, and there is still no established approach. The newer spec says this:
> It’s NOT RECOMMENDED to record the same exception more than once. It’s NOT RECOMMENDED to record exceptions that are handled by the instrumented library.

But it is not clear how to actually implement this in real code when the official way of recording exceptions is record -> rethrow (https://opentelemetry.io/docs/specs/otel/trace/exceptions/#recording-an-exception):
```
Span span = myTracer.startSpan(/*...*/);
try {
  // Code that does the actual work which the Span represents
} catch (Throwable e) {
  span.recordException(e);
  span.setAttribute(AttributeKey.stringKey("error.type"), e.getClass().getCanonicalName())
  span.setStatus(StatusCode.ERROR, e.getMessage());
  throw e;
} finally {
  span.end();
}
```
This is what we do in zio-telemetry as well as other ecosystems such as python, C#, rust, and so on. 

Also, while I was investigating this, I encountered this article on the official OTEL blog: https://opentelemetry.io/blog/2026/deprecating-span-events/

So eventually we would need to switch to log records instead of the `.recordException` API. But the above issue remains the same and it is kind of unsolvable. It is up to an end user of the OTEL API to decide on when and where to record their exceptions.
If you really want to avoid duplicating them, you need to have separate handlers/status mappers. One solution I came up with is to record exceptions in the top-level error handler such as an http error handler. This approach makes it difficult to recognise exactly where the exception was thrown, though. There is always a trade-off :) 

Thanks for listening! :)

dontgitit:
It definitely feels like the best place for the exception to show up is in the span where it happened.
I have mixed feelings about "don't record the same exception multiple times"... on one hand, it's not terribly helpful, but on the other... I think it might be? for example, consider
```scala
def a() = {
  tracer.span("a") { _ => 
    b().ignore() 
  }
}

def b() = {
  tracer.span("b") { _ =>
    ZIO.foreachPar(1 to 10)(c(_)).flatMap { _ =>
      ZIO.fail(new RuntimeException("something else"))
    }
  }
}

def c(i: Int) = {
  tracer.span("c") { _ =>
    if Random.nextInt(100) < i then
      ZIO.fail(new RuntimeException(s"boom: $i"))
    else 
      ZIO.unit
  }
}
```

Here, the span for `c` should definitely record the exception. But what about the span for `b`? it's going to fail with the same exception as `c`... if we just set span ERROR status but don't record the exception, it becomes unclear how `b` failed. This is compounded if `b` calls a handful of methods and each of those can fail, or maybe something like `foreachPar` (actually not sure if that returns the "first" exception or all of them)... 

I think in general having the full exception on each span where the exception wasn't handled is helpful!

PR #1226, opened 2026-08-20
refactor: remove duplicated default arguments from Tracing implementations

khajavi:
## What

Default argument values were declared twice — once on the `Tracing` trait, once again on the anonymous implementation inside `Tracing.scoped` — in both the `opentelemetry` and `opencensus` modules. This removes the second copy: 23 duplicated defaults in `opentelemetry/.../tracing/Tracing.scala` and 12 in `opencensus/.../Tracing.scala`.

Reported by @dontgitit on [Discord](https://discord.com/channels/629491597070827530/639825316021272602/1539740044858892339).

## Why

Two reasons beyond DRY.

**1. The copies are unnecessary.** An override that omits defaults inherits them from the overridden method. The compiler attaches the synthetic default getters (`span$default$2` and friends) to `Tracing` itself, so callers keep working whether they hold the trait type or the concrete one.

**2. Duplicated defaults are a silent hazard.** Those getters are ordinary virtual methods, so an implementation's default *overrides* the trait's — even for callers typed as the trait:

```scala
trait T { def g(x: Int = 1): Int }
object Impl extends T { override def g(x: Int = 99): Int = x }

val t: T = Impl
t.g() // 99, not 1
```

If the copies ever drift, the trait's documented default becomes a lie and nothing fails to compile. `spanScoped` had already half-drifted — four parameters, only `statusMapper` still carrying a default — which is exactly the accident this prevents.

Minor bonus: fewer synthetic `$default$n` definitions on the implementations.

Left alone deliberately: defaults on factory methods (`Tracing.live`, `Tracing.scoped`) and on the concrete `aspects.*` helpers. Those are the API surface, not overrides.

## Compatibility

Source-compatible: the trait defaults are untouched, and the anonymous classes are unnameable by users — `live`/`scoped` return `Tracing`.

Binary-compatible: verified against the compiled 2.13 output. The anon impl class still exposes all 26 `$default$n` getters, now as mixin forwarders to the trait defaults. `javap` member dumps before and after the change are byte-for-byte identical.

No behavior change: every removed default was value-identical to its trait counterpart (`SpanKind.INTERNAL`, `Attributes.empty()`, `StatusMapper.default`, `Seq.empty`; OpenCensus: `Span.Kind.SERVER`, `ErrorMapper.default`, `Map.empty`).

## Verification

- `+Test/compile` green on 2.12.21, 2.13.18, 3.3.8
- `scalafmtCheckAll` green
- `javap` pre/post member-list diff on `Tracing$$anon$5`: identical

## Note, not addressed here

`project/MimaSettings.scala` currently accepts a `failOnProblem: Boolean` parameter and then ignores it, hardcoding `mimaFailOnProblem := false`, while all four call sites pass `true`. Combined with `exclude[Problem]` filters covering every shipping package, the Mima check cannot fail. Out of scope for this PR, but worth a look.

### DIFF
--- a/opencensus/src/main/scala/zio/telemetry/opencensus/Tracing.scala
+++ b/opencensus/src/main/scala/zio/telemetry/opencensus/Tracing.scala
@@ -119,9 +119,9 @@ object Tracing {
 
         override def span[R, E, E1 <: E, A](
           name: String,
-          kind: Span.Kind = Span.Kind.SERVER,
-          toErrorStatus: ErrorMapper[E] = ErrorMapper.default,
-          attributes: Map[String, AttributeValue] = Map.empty
+          kind: Span.Kind,
+          toErrorStatus: ErrorMapper[E],
+          attributes: Map[String, AttributeValue]
         )(zio: => ZIO[R, E1, A])(implicit trace: Trace): ZIO[R, E1, A] =
           ZIO.scoped[R] {
             for {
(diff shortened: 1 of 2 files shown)


dontgitit:
FWIW i think there are similar patterns in `test` files, such as https://github.com/zio/zio-telemetry/pull/1223/changes#diff-15bd821e5f1ce1ac876da106c4c78db335e0da9bffed997c6b4ae6e9632d4e4fR76-R102

grouzen:
Could you please do it in the v4.0.0-rc branch, where all the new development is happening? I'm merging the dependency updates into series/2.x only for now. It would be difficult to maintain two branches and resolve conflicts since they diverge drastically at this point.

PR #1236, opened 2026-09-02
fix(opentelemetry): record causes as exceptions in logs

hdriviere:
The OpenTelemetry `ZLogger` receives a `Cause[Any]`, but currently ignores it when emitting the log record. As a result, logs produced with `ZIO.logErrorCause` omit the standard OpenTelemetry exception attributes:

- `exception.type`
- `exception.message`
- `exception.stacktrace`

This change passes non-empty causes to `LogRecordBuilder.setException` and handles the different ZIO cause shapes explicitly:

- A single typed `Throwable` keeps its original exception details.
- A single defect keeps its original exception details.
- Composite or interrupted causes use `FiberFailure`, preserving the complete ZIO cause.

Regression tests cover a typed exception, a defect, and a composite cause. The composite test verifies that both exceptions remain present in `exception.stacktrace`.

Tested with Scala 2.12, 2.13, and 3.


grouzen (approved):
Great job! Thanks you for your contribution! I'll try to cut next RC version ASAP!

